PR #56, opened 2019-07-09
[kustomize] kustomize edit fix

binoue:
As `kustomize edit fix` command migrates a kustomization file working with previous versions to one working with 2.0.0.

See the link below.
https://github.com/kubernetes-sigs/kustomize/blob/master/docs/v2.0.0.md#compatible-changes-new-features

Unfortunately, the command automatically edits the original kustomization file and applys without any warning and such behavior is not appropriate in CI test.

Therefore, in this pull request, the following changes were made.
- run `kustomize edit fix` command to all kustomization.yaml files to work with 2.0.0
- delete `kustomize edit fix` script to prevent from working in CI test

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -26,10 +26,6 @@ test: install.yaml validation
 kustomize-check:
 	./kustomize-check.sh
 
-# NOTE: Enable this target when Argo CD uses kustomize 2.0.0
-#kustomize-fix:
-#	./kustomize-fix.sh
-
 setup:
 	curl -sSLf -O https://github.com/kubernetes-sigs/kustomize/releases/download/v$(KUSTOMIZE_VERSION)/kustomize_$(KUSTOMIZE_VERSION)_linux_amd64
 	sudo mv kustomize_$(KUSTOMIZE_VERSION)_linux_amd64 /usr/local/bin/kustomize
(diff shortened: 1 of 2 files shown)


morimoto-cybozu:
`kustomize edit fix` seems to have side effects:
* It removes indents of list items.
  Before fix:
  ```
  foo:
    - 1
    - 2
  ```
  After fix:
  ```
  foo:
  - 1
  - 2
  ```
* It sorts map keys.
  Before fix:
  ```
  - name: alertmanager
    behavior: merge
    files:
    - alertmanager/env.template
  ```
  After fix:
  ```
  - behavior: merge
    files:
    - alertmanager/env.template
    name: alertmanager
  ```

Especially key sort breaks readability of kustomization YAMLs.

I suggest using `kustomize edit fix` only for detection of incompatibility.
In our case, `kustomize edit fix` produced no meaningful diff in terms of compatibility.
We don't need to modify the current YAMLs, do we?

binoue:
After some discussion, as stated in the previous comment, using `kustomize edit fix` command only breaks the readability and gives no good benefit.

Therefore, this pull request reverts  `03bd04b` commit and use the original `kustomization.yaml` files

PR #58, opened 2019-07-11
Align to recommended label key

dulltz:


test/network-policy_test.go
```diff
@@ -32,11 +32,11 @@ spec:
   replicas: 2
   selector:
     matchLabels:
-      run: testhttpd
+      app.kubernetes.io/name: testhttpd
   template:
     metadata:
       labels:
-        run: testhttpd
+        app.kubernetes.io/name: testhttpd
```
kfyharukz: Please confirm following selectors on Service and NetworkPolicy manifests.

PR #61, opened 2019-07-16
[teleport] deploy teleoprt

zoetrope:


binoue:
As creating certificates needs time, it may be required to add `--timeout` option when running `argocd app sync` command.

PR #93, opened 2019-08-13
Fix monitoring stack

dulltz:
- Allow prometheus access to cke-etcd's metrics
- Allow prometheus access to felix's metrics
- Fix node-exporter-full.json for provisioning Grafana
- Request 10Gi PV for Grafana and Prometheus on staging environment
- Relaxed etcd alert rules threshold
- Added alert rule using `up` metrics
- Enable to switch etcd cluster on the dashboard

### DIFF
--- a/monitoring/base/prometheus/alert_rules.yaml
+++ b/monitoring/base/prometheus/alert_rules.yaml
@@ -51,3 +51,13 @@ groups:
         annotations:
           summary: "{{ $labels.instance }}, {{ $labels.job }} of etcd DB space increases {{ $value | humanize }}B/h"
           runbook: "Please consider to find root causes, and solve the problems"
+      - alert: CKEEtcdMissing
+        expr: absent(up{job="cke-etcd"})
+        labels:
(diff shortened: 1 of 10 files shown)


monitoring/base/prometheus/alert_rules.yaml
```diff
@@ -44,10 +44,20 @@ groups:
...
+      - alert: CKEEtcdMissing
+        expr: absent(up{job="cke-etcd"})
+        labels:
+          severity: warning
+        for: 10m
+      - alert: BootserverEtcdMissing
+        expr: absent(up{job="bootserver-etcd"})
+        labels:
+          severity: warning
+        for: 10m
```
mitsutaka: Add more alert about `up == 0` for any metrics are down.

dulltz: I'll add the rule, but comment it out because we don't want to apply it now.

dulltz:
Updated

PR #95, opened 2019-08-15
[gatekeeper] replace opa to gatekeeper

zoetrope:
OpenPolicyAgent and kube-mgmt were reborn as Gatekeeper.
Gatekeeper has several benefits.

* We can create policy as Custom Resources.
* Kubernetes data can be replicated without restarting the container.
* We can periodically apply policies.


morimoto-cybozu:
Some manual operations are necessary before merging and applying this.
Please make document for apply operation.

PR #98, opened 2019-08-16
[grafana][prometheus] Fix monitoring settings

dulltz:
- Allow Prometheus to scrape Contour's metrics.
- Fix default time range of Calico dashboard.
- Disable excessive labelmap.
- [stage0] Resize the storage of Prometheus 10Gi to 150Gi

### DIFF
--- a/monitoring/base/prometheus/prometheus.yaml
+++ b/monitoring/base/prometheus/prometheus.yaml
@@ -158,6 +158,17 @@ scrape_configs:
         target_label: __address__
       - action: labelmap
         regex: __meta_kubernetes_endpoint_(.+)
+  - job_name: "contour"
+    kubernetes_sd_configs:
+      - role: service
+        namespaces:
+          names: ["ingress"]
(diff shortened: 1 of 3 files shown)


monitoring/base/prometheus/prometheus.yaml
```diff
@@ -158,6 +158,17 @@ scrape_configs:
...
+    kubernetes_sd_configs:
+      - role: service
+        namespaces:
+          names: ["ingress"]
+    relabel_configs:
+      - source_labels: [__meta_kubernetes_service_name]
+        action: keep
+        regex: contour-metrics
+      - action: labelmap
+        regex: __meta_kubernetes_service_(.+)
```
mitsutaka: It seems too much target labels generated. It should be disabled now. At least `job` and IP address related label should be included.

dulltz: 🆗

PR #108, opened 2019-09-03
[argocd] Add missing empty ConfigMaps

mitsutaka:


mitsutaka:
This commit will be finally included in https://github.com/cybozu-go/neco-apps/pull/110

PR #114, opened 2019-09-06
[monitoring] Enable to scrape Prometheus and Alertmanager

dulltz:
- https://github.com/cybozu-go/neco/issues/527 Use `.Annotations` instead of `.CommonAnnotations` for cases without summary annotation 
- Enable to scrape prometheus and alertmanager pods 
- Rename machines-endpoints targets to node-exporter targets

### DIFF
--- a/monitoring/base/prometheus/prometheus.yaml
+++ b/monitoring/base/prometheus/prometheus.yaml
@@ -29,13 +29,13 @@ scrape_configs:
         namespaces:
           names: ["monitoring"]
     relabel_configs:
-      - source_labels: [__meta_kubernetes_pod_label_app_kubernetes_io_name]
+      - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
         action: keep
         regex: alertmanager
       - source_labels: [__address__]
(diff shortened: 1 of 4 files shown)


monitoring/base/prometheus/prometheus.yaml
```diff
@@ -9,6 +9,34 @@ alerting:
...
+        namespaces:
+          names: ["monitoring"]
+    relabel_configs:
+      - source_labels: [__meta_kubernetes_pod_label_app_kubernetes_io_name]
+        action: keep
+        regex: prometheus
+      - source_labels: [__address__]
+        action: replace
+        regex: ([^:]+)(?::\d+)?
+        replacement: ${1}:9090
```
masa213f: I want to avoid hard coding of port numbers in this file.
Could you add "prometheus.io/port" annotation to pods?

dulltz: OK

dulltz: Oops, I forgot to fix alertmanager

PR #130, opened 2019-09-19
[kindtest] stop using Cloud DNS

dulltz:
## Purpose

Stop using GCP service account file (account.json) in kindtest environment.

## Details
 
- Use self-signed-issuer instead of cloud-dns issuer at kindtest.
- Make external-dns watch IngressRoute directly, so delete DNSEndpoint resources and its definition.
- Stop deploying external-dns at kindtest because it is not used.


test/contour_test.go
```diff
@@ -181,60 +181,32 @@ spec:
...
 
-		By("accessing with curl: https")
-		if withKind {
-			ExecSafeAt("", "curl", "-sfL", "-o", "lets.crt", "https://letsencrypt.org/certs/fakelerootx1.pem")
-		} else {
-			ExecSafeAt(boot0, "HTTPS_PROXY=http://10.0.49.3:3128",
-				"curl", "-sfL", "-o", "lets.crt", "https://letsencrypt.org/certs/fakelerootx1.pem")
+		if !withKind {
+			By("accessing with curl: https")
+			if withKind {
```
zoetrope: This condition is not necessary.
withKind will be always false.

PR #136, opened 2019-09-24
[monitoring] scrape kube-state-metrics's own metrics

dulltz:
Fix the missing configuration of Prometheus for scraping kube-state-metrics's own metrics.
- The port is configured by [`--telemetry-port`](https://github.com/cybozu/neco-containers/blob/master/kube-state-metrics/Dockerfile#L22).
- The port is already specified in [kube-state-metrics](https://github.com/cybozu-go/neco-apps/blob/master/monitoring/base/kube-state-metrics/service.yaml#L16-L18) service.

### DIFF
--- a/monitoring/base/prometheus/prometheus.yaml
+++ b/monitoring/base/prometheus/prometheus.yaml
@@ -90,7 +90,7 @@ scrape_configs:
         regex: "kube-state-metrics"
       - source_labels: [__meta_kubernetes_service_port_name]
         action: keep
-        regex: http-metrics
+        regex: (http-metrics|telemetry)
   - job_name: 'node-exporter'
     kubernetes_sd_configs:
       - role: endpoints


monitoring/base/prometheus/prometheus.yaml
```diff
@@ -70,6 +70,9 @@ scrape_configs:
       - action: labelmap
         regex: __meta_kubernetes_node_label_sabakan_cke_cybozu_com_(.+)
         replacement: sabakan_${1}
+      - action: labelmap
+        regex: __meta_kubernetes_node_annotation_topolvm_cybozu_com_(.+)
+        replacement: topolvm_${1}
```
zoetrope: I think `topolvm_capacity` is high cardinality.
Should it be collected as a label?

dulltz: What you are saying is correct. I'll stop mapping this annotation to the target label

PR #151, opened 2019-09-30
Add Prometheus rules for k8s objects

dulltz:
- Add alert rules using `up` metrics.
- Add alert rules for the following kind objects.
    - CronJob
    - DaemonSet
    - Deployment
    - Job
    - Node
    - Pod
    - StatefulSet
- Add some recording rules about kube-apiserver.
- Add unit tests for the above alert rules.
- This PR refers to [this examples](https://github.com/coreos/kube-prometheus/blob/f458e85e5d7675f7bc253072e1b4c8892b51af0f/manifests/prometheus-rules.yaml), so insert the license and notice it at README


zoetrope:
Please check apache license 2.0.

> You must cause any modified files to carry prominent notices stating that You changed the files

> If the Work includes a "NOTICE" text file as part of its distribution, then any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works, 

`kube-prometheus` has a NOTICE file.
https://github.com/coreos/kube-prometheus/blob/master/NOTICE

PR #160, opened 2019-10-02
[monitoring] Add readiness/liveness probe for Prometheus

dulltz:
refer to the link https://github.com/helm/charts/blob/f5bcdc1c4e6c039cfd28184f2ed00f0f38470771/stable/prometheus/templates/server-deployment.yaml#L97-L108

### DIFF
--- a/monitoring/base/prometheus/statefulset.yaml
+++ b/monitoring/base/prometheus/statefulset.yaml
@@ -27,6 +27,18 @@ spec:
             - --web.enable-admin-api
           ports:
             - containerPort: 9090
+          readinessProbe:
+            httpGet:
+              path: http://prometheus:9090/-/ready
+              port: 9090
+            initialDelaySeconds: 30
+            timeoutSeconds: 30
+          livenessProbe:
+            httpGet:
+              path: http://prometheus:9090/-/healthy
+              port: 9090
+            initialDelaySeconds: 30
+            timeoutSeconds: 30
           volumeMounts:
             - name: prometheus-config-volume
               mountPath: /etc/prometheus/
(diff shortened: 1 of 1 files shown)


monitoring/base/prometheus/statefulset.yaml
```diff
@@ -27,6 +27,18 @@ spec:
             - --web.enable-admin-api
           ports:
             - containerPort: 9090
+          readinessProbe:
+            httpGet:
+              path: http://prometheus:9090/-/ready
```
zoetrope:
```suggestion
              path: /-/ready
```

monitoring/base/prometheus/statefulset.yaml
```diff
@@ -27,6 +27,18 @@ spec:
...
+              port: 9090
+            initialDelaySeconds: 30
+            timeoutSeconds: 30
+          livenessProbe:
+            httpGet:
+              path: http://prometheus:9090/-/healthy
```
zoetrope:
```suggestion
              path: /-/healthy
```

monitoring/base/prometheus/statefulset.yaml
```diff
@@ -27,6 +27,18 @@ spec:
             - --web.enable-admin-api
           ports:
             - containerPort: 9090
+          readinessProbe:
+            httpGet:
+              path: http://prometheus:9090/-/ready
+              port: 9090
+            initialDelaySeconds: 30
```
zoetrope: Is it enough?
It takes 10 minutes to recover from a crash.

dulltz: That's true! I'll increase `readinessProbe.failureThreshold`.

PR #163, opened 2019-10-03
CI: delete gcp0.dev-ne.co DNS records daily

ymmt2005:


.circleci/config.yml
```diff
@@ -247,6 +247,19 @@ jobs:
...
+      - image: google/cloud-sdk
+    steps:
+      - run:
+          name: Store Service Account
+          command: |
+            echo $GCLOUD_SERVICE_ACCOUNT > account.json
+            gcloud auth activate-service-account --key-file=account.json
+      - run:
+          name: Delete DNS records
+          command: gcloud --project=neco-test dns record-sets import --zone=gcp0 --delete-all-existing /dev/null
```
mitsutaka: NOTE: https://stackoverflow.com/questions/38636010/removing-dns-entries-with-gcloud

ymmt2005: Good catch :+1:

PR #180, opened 2019-10-10
Upgrade TopoLVM to 0.2.0

dulltz:
Update TopoLVM's manifests to v0.2.0 according to the following samples.
https://github.com/cybozu-go/topolvm/tree/v0.2.0/deploy

And implement the upgrade test.


topolvm/base/crd.yaml
```diff
@@ -61,7 +64,14 @@ spec:
...
-    storage: true
+    - name: v1
+      served: true
+      storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
```
ymmt2005: Argo CD cannot handle CRD status properly.  So remove these lines.

topolvm/base/scheduler.yaml
```diff
@@ -71,5 +71,5 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
               - matchExpressions:
-                  - key: cke.cybozu.com/master
+                  - key: node-role.kubernetes.io/master
```
ymmt2005: This change is wrong.  In Neco cluster, control plane nodes are
labeled with `cke.cybozu.com/master`.

masa213f: Fixed it. 47fd967

topolvm/base/crd.yaml
masa213f: Fixed it. 47fd967

PR #215, opened 2019-11-02
Add internet-egress namespace

ymmt2005:
internet-egress namespace is used for Squid and Unbound
to connect to the Internet.  To apply network policy w/o
restriction enforced by neco-admission, the namespace
need to have label "control-plane: true".


ymmt2005:
@binoue 
We did the same thing for `kube-system` namespace already.
Of course, it is not deleted.

PR #221, opened 2019-11-06
ArgoCD RBAC for maneki

masa213f:
- Add an ArgoCD project for Maneki team.
- Permit just the Neco team to access the ArgoCD default project.

### DIFF
--- a/team-management/base/maneki/project.yaml
+++ b/team-management/base/maneki/project.yaml
@@ -1,0 +1,26 @@
+apiVersion: argoproj.io/v1alpha1
+kind: AppProject
+metadata:
+  name: maneki
+  namespace: argocd
+spec:
+  sourceRepos:
+  - '*'
+  destinations:
+  - namespace: maneki
+    server: https://kubernetes.default.svc
+  namespaceResourceBlacklist:
+  - group: ''
+    kind: ResourceQuota
+  - group: ''
+    kind: LimitRange
+  - group: ''
+    kind: NetworkPolicy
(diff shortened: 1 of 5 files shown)


team-management/base/maneki/project.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: argoproj.io/v1alpha1
+kind: AppProject
+metadata:
+  name: maneki
+  namespace: argocd
+spec:
+  sourceRepos:
+  - '*'
+  destinations:
+  - namespace: maneki
```
zoetrope: I think `sandbox` namespace is necessary.

team-management/base/maneki/project.yaml
```diff
@@ -0,0 +1,26 @@
...
+    server: https://kubernetes.default.svc
+  namespaceResourceBlacklist:
+  - group: ''
+    kind: ResourceQuota
+  - group: ''
+    kind: LimitRange
+  - group: ''
+    kind: NetworkPolicy
```
zoetrope: The group of `NetworkPolicy` is `extensions` and `networking.k8s.io`.
Do you need to specify it?

team-management/base/maneki/project.yaml
masa213f: Yes!

team-management/base/maneki/project.yaml
masa213f: We need `networking.k8s.io`!

PR #225, opened 2019-11-08
Wait squid for network-policy test

kna:


test/network-policy_test.go
```diff
@@ -216,11 +216,23 @@ spec:
...
+			}
+
+			if deployment.Status.ReadyReplicas != 2 {
+				return fmt.Errorf("squid deployment's ReadyReplicas is not 2: %d", int(deployment.Status.ReadyReplicas))
+			}
+			if deployment.Status.UpdatedReplicas != 2 {
+				return fmt.Errorf("squid deployment's UpdatedReplicas is not 2: %d", int(deployment.Status.UpdatedReplicas))
+			}
 
 			stdout, stderr, err = ExecAt(boot0, "kubectl", "exec", "-n=internet-egress", podName, "-c", "ubuntu", "true")
```
ymmt2005: Now that podName is not set, this test can/should be removed.

test/network-policy_test.go
```diff
@@ -216,11 +216,23 @@ spec:
 		var podName string
```
ymmt2005: get rid of this.

PR #226, opened 2019-11-08
Cert-manager: Upgrade to 0.11.0

kna:
- Update cert-manager to v0.11.0
- Update contour-plus to v0.3.0
- Add CRDs for contour v1.0.0 and cert-manager v0.11.0
  - Clone the YAML files from upstream
  - Add patch file to remove status fields by JSON patch


go.mod
```diff
@@ -1,8 +1,11 @@
 module github.com/cybozu-go/neco-apps
 
-go 1.12
+go 1.13
 
-replace launchpad.net/gocheck => github.com/go-check/check v0.0.0-20180628173108-788fd7840127
+replace (
+	k8s.io/client-go => k8s.io/client-go v0.0.0-20190805141520-2fe0317bcee0
```
ymmt2005: why is this line needed?

llamerada-jp: go.mod in client-go is not using a policy of the new declaration format.
And go.mod in cert-manager v.0.11.0 did not consider for old declaration format of client-go.
We need to specify an explicit version of client-go to avoid build error from the wrong version of it.

PR #227, opened 2019-11-08
update manifests for ECK 1.0.0-beta.1

zoetrope:
* Use JSONPath to make it easier for us to maintain manifests to keep up with upstream.
* Add PodSecurityPolicy to avoid permission denied. (ref: https://github.com/elastic/cloud-on-k8s/issues/2108 )


test/setup_test.go
```diff
@@ -225,12 +225,24 @@ func testSetup() {
 		ExecSafeAt(boot0, "cd neco-apps; git checkout "+commitID)
 	})
 
+	if doUpgrade {
```
binoue: If this code is temporal and expected to be deleted after this commit is merged to stage and release branch, please add comments about that.

PR #238, opened 2019-11-28
Contour 1.0

dulltz:
- Upgrade contour to v1.0
- Use `HTTPProxy` instead of `IngressRoute`
- Create the OTS patch
  - Note that some manifests does not inherit the upstream


ingress/base/contour/deployment.yaml
```diff
@@ -4,127 +4,105 @@ metadata:
   labels:
     app.kubernetes.io/name: contour
   name: contour
-  namespace: ingress
+  namespace: projectcontour
 spec:
+  strategy:
```
ymmt2005: seems an invalid YAML?

ingress/base/contour/deployment.yaml
```diff
@@ -4,127 +4,105 @@ metadata:
...
   selector:
     matchLabels:
       app.kubernetes.io/name: contour
-  replicas: 2
```
ymmt2005: Why do you remove this?

ingress/base/contour/contour-config.yaml
```diff
@@ -1,18 +1,16 @@
...
-    # disablePermitInsecure: false
+    disablePermitInsecure: false
     tls:
-      # minimum TLS version that Contour will negotiate
       minimumProtocolVersion: "1.2"
+    leaderelection:
+      configmap-name: contour-leader
+      configmap-namespace: ingress
+    accesslog-format: envoy
+
```
ymmt2005: garbage line

ingress/base/contour/deployment.yaml
ymmt2005: The off-the-shelf strategy seems good.

ymmt2005:
Please change the PR so that it does not patch deployments/daemonsets but instead
replace the whole manifests of deployments/daemonsets entirely with ours.

ingress/base/contour/deployment.yaml
kfyharukz: I wrote the same strategy in the upstream.
Also, I added the strategy to envoy deployment (it is different from the upstream, but I think it should be written.)

docs/maintenance.md
```diff
@@ -97,6 +97,9 @@ cd contour
 git diff vA.B.C...vX.Y.Z examples/contour
 ```
 
+Note that not all manifests inherit the upstream (e.g. contour and envoy deployments).
+Please check `kustomization.yaml` which manifest inherits or not.
+
```
ymmt2005: Please add these descriptions:

- We do not use contour's certificate issuance feature, but use cert-manager to issue certificates required for gRPC.
- We change Envoy from DaemonSet to Deployment.

kfyharukz: Thanks your comments. I reflected them.

PR #241, opened 2019-12-02
Recreate Certificate if CertificateRequest fails

binoue:
When CertificateRequest fails, `cert-manager` does not retry in 1h.
See [this](https://github.com/jetstack/cert-manager/blob/v0.11.1/pkg/controller/certificates/sync.go#L410-L440)

However, it is too long for our tests and `neco-apps` CIs sometimes fail.
To fix this, this pull request made the following changes.
- recreate Certificate when its CertificateRequest fails
- delete Certificate when `contour-plus` failed to succeed its CertificateRequest.

### DIFF
--- a/test/cert-manager_test.go
+++ b/test/cert-manager_test.go
@@ -112,6 +112,18 @@ spec:
 
 			for _, st := range cert.Status.Conditions {
 				if st.Type != certmanagerv1alpha2.CertificateConditionReady {
+					// Check if recreate Certificate resource is necessary
+					failed, err := isCertificateRequestFailed(cert)
+					if err != nil {
+						return err
+					}
+					if failed {
+						err = recreateCertificate("test-certificate", "cert-manager", certificate)
+						if err != nil {
+							return err
+						}
+						return fmt.Errorf("recreate Certificate")
+					}
 					continue
 				}
 				if st.Status == "True" {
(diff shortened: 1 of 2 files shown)


test/cert-manager_test.go
```diff
@@ -126,3 +139,55 @@ spec:
...
+		}
+	}
+
+	if targetCertReq.Name == "" {
+		return false, nil
+	}
+
+	reason := targetCertReq.Status.Conditions[0].Reason
+	if reason == "Failed" {
+		log.Error("CertificateRequest failed", map[string]interface{}{
```
zoetrope: Where are the logs output?
I think you should output logs to GinkgoWriter or stdout.

test/cert-manager_test.go
```diff
@@ -126,3 +139,55 @@ spec:
...
+	for _, cr := range certReqList.Items {
+		for _, or := range cr.OwnerReferences {
+			if or.Name == cert.Name {
+				targetCertReq = cr
+				break OUTER
+			}
+		}
+	}
+
+	if targetCertReq.Name == "" {
```
zoetrope: Can you confirm it in the above loop?

zoetrope: Is this a confirmation that the target does not exist?

binoue: Yes. I think there is a possibility that CertificateRequest does not exist because of checking CertificateRequest too early.

zoetrope: I think `targetCertReq` should be a pointer.

test/contour_test.go
```diff
@@ -221,6 +222,46 @@ spec:
...
+				if err != nil {
+					return err
+				}
+
+				if len(cert.Status.Conditions) == 0 {
+					return errors.New("status not found")
+				}
+
+				for _, st := range cert.Status.Conditions {
+					if st.Type != certmanagerv1alpha2.CertificateConditionReady {
```
zoetrope: Is it correct?

Status type is only `Ready`.
https://github.com/jetstack/cert-manager/blob/0f99bcdeacefcc3dad40254583734ec0b1c08a7e/pkg/apis/certmanager/v1alpha2/types_certificate.go#L196

test/cert-manager_test.go
zoetrope: no problem. :smile:

test/contour_test.go
binoue: This comparison is useless. Fix the condition.

test/cert-manager_test.go
```diff
@@ -111,12 +112,21 @@ spec:
 			}
 
 			for _, st := range cert.Status.Conditions {
-				if st.Type != certmanagerv1alpha2.CertificateConditionReady {
```
ymmt2005: In order to prepare the future status types, this should not be removed.

PR #257, opened 2019-12-09
Fix recreating Certificate bug

binoue:


test/cert-manager_test.go
```diff
@@ -167,15 +167,24 @@ OUTER:
...
-	reason := targetCertReq.Status.Conditions[0].Reason
-	if reason == "Failed" {
-		log.Error("CertificateRequest failed", map[string]interface{}{
-			"certificate name":         cert.Name,
-			"certificate request name": targetCertReq.Name,
-		})
-		return true, nil
+	if len(targetCertReq.Status.Conditions) == 0 {
+		return false, errors.New("status not found")
 	}
```
ymmt2005: don't need.

test/cert-manager_test.go
```diff
@@ -167,15 +167,24 @@ OUTER:
...
-	return false, fmt.Errorf("CertificateRequest is not ready")
+
+	for _, st := range targetCertReq.Status.Conditions {
+		if st.Type != certmanagerv1alpha2.CertificateRequestConditionReady {
+			continue
+		}
+		reason := st.Reason
+		if reason == "Failed" {
```
ymmt2005: Why don't you reference `Status`?
https://godoc.org/github.com/jetstack/cert-manager/pkg/apis/meta/v1#ConditionStatus

binoue: As shown in [here](https://cert-manager.io/docs/concepts/certificaterequest/#conditions), status can be False when the reason is InProgress. Thinking of that, I thought it is better to  recreate Certificate only when its reason is Failed.

# The above document column "Ready" is [this](https://github.com/jetstack/cert-manager/blob/abb680756e84b56df47f85a8ce732a0e09e0f171/deploy/manifests/00-crds.yaml#L3-L12)

The possible reason values are as follows
```
const (
	CertificateRequestReasonPending = "Pending"
	CertificateRequestReasonFailed  = "Failed"
	CertificateRequestReasonIssued  = "Issued"
)
```
https://github.com/jetstack/cert-manager/blob/master/pkg/apis/certmanager/v1alpha2/types_certificaterequest.go#L25-L29

ymmt2005: So, why don't you use that constant?

test/cert-manager_test.go
ymmt2005: not fixed yet.

binoue: Sorry.. I did not push my local commit..

test/cert-manager_test.go
```diff
@@ -167,15 +163,19 @@ OUTER:
...
-		log.Error("CertificateRequest failed", map[string]interface{}{
-			"certificate name":         cert.Name,
-			"certificate request name": targetCertReq.Name,
-		})
-		return true, nil
+	for _, st := range targetCertReq.Status.Conditions {
+		if st.Type != certmanagerv1alpha2.CertificateRequestConditionReady {
+			continue
+		}
+		if st.Status == certmanagerv1alpha2.CertificateRequestReasonFailed {
```
ymmt2005: `st.Reason`

PR #258, opened 2019-12-09
Stop specifying fields by number

dulltz:
Use strategic merge patch

Signed-off-by: dulltz <isrgnoe@gmail.com>


elastic/base/statefulset.yaml
```diff
@@ -0,0 +1,24 @@
...
+metadata:
+  name: elastic-operator
+  namespace: elastic-system
+  labels:
+    app.kubernetes.io/name: elastic-operator
+spec:
+  selector:
+    matchLabels:
+        app.kubernetes.io/name: elastic-operator
+        control-plane: elastic-operator
```
morimoto-cybozu: This label is specified in the upstream manifest, so redundant here.

Use indent of 2 spaces.

PR #269, opened 2019-12-17
ingress: Divide Contour Deployment according to Services

morimoto-cybozu:
This PR divides a Deployment of `contour` into three: for the Internet, for the Forest network, and for the bastion network.
Now an HTTPProxy object should have an annotation of `kubernetes.io/ingress.class: {global|forest|bastion}` to indicate the network it wants to use.


ingress/base/common/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - 02-rbac.yaml
```
ymmt2005: There is no reason to name this `02-rbac.yaml`.

PR #276, opened 2019-12-20
[CI] Release release-2019.12.20-6377

cybozu-neco:


ymmt2005:
After this is merged, the following operation needs to be run:

```console
$ argocd app sync namespaces --prune
```

PR #302, opened 2020-01-06
Add recording rules for scheduler

dulltz:
- Add recording rules for scheduler
- Add health check alert rules for scheduler and controller-manager


morimoto-cybozu:
This depends on network policy updates in https://github.com/cybozu-go/neco-apps/pull/303 .

PR #303, opened 2020-01-06
prometheus: resize PV for dctest

dulltz:
dctest failed due to the disk size of Prometheus, so resize disk from 1GB to 5GB.
https://circleci.com/gh/cybozu-go/neco-apps/6599?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-l


monitoring/base/prometheus/statefulset.yaml
```diff
@@ -61,4 +61,4 @@ spec:
         storageClassName: topolvm-provisioner
         resources:
           requests:
-            storage: 1Gi
+            storage: 5Gi
```
kfyharukz: should write prod env settings as base,
and write in overlays for other environments.

dulltz: I agree it.
However I keep the overlays of `spec.template.spec.cotainers[].resources` because it is possible to change the default value by `LimitRange`.

PR #310, opened 2020-01-09
Add local-pv-provisioner app

masa213f:
Add manifests for local-pv-provisioner.


local-pv-provisioner/base/daemonset.yaml
```diff
@@ -0,0 +1,47 @@
...
+        - name: dev
+          hostPath:
+            path: /dev/crypt-disk/by-path
+      nodeSelector:
+        cke.cybozu.com/role: ss
+      tolerations:
+        - key: cke.cybozu.com/role
+          operator: Equal
+          value: storage
+          effect: NoExecute
```
ymmt2005: Can be omitted.

local-pv-provisioner/base/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: manager-rolebinding
```
zoetrope: I think this name should be changed. (e.g. `local-pv-provisioner-rolebinding`)
Because ClusterRoleBinding is a cluster resource.

llamerada-jp: changed the name to `local-pv-provisioner`. 
I did not add the postfix of `-rolebinding` because I checked the staging environment and found that it was a minority.

PR #327, opened 2020-01-10
[monitoring] Add alert rules for the applications.

kfyharukz:
Add alert rules for applications on Kubernetes.
The generic rules are already exists in neco-apps, so this PR introduces application specific alert rules.

In addition, set the policy for the `severity` label.

---

#### external-dns
- Alert if registry has any errors
- Alert if dns endpoints have any errors

#### cert-manager
- Alert if the status of Certificate is not Ready

#### contour
- Alert if the count of invalid httpproxy resources is increasing

#### metallb
- Alert if any BGP sessions are down
- Alert if configurations are stale
- Alert if the address pool will be exhausted

#### network policy
- Alert if the time to apply data plane become slowly
- Alert if the count of errors in data plane is increasing

#### node-exporter
- Alert if the count of conntrack entries will be exhausted

#### topolvm
- Alert if the disk size in the whole of cluster will be exhausted

### DIFF
--- a/monitoring/base/prometheus/alert_rules/topolvm.yaml
+++ b/monitoring/base/prometheus/alert_rules/topolvm.yaml
@@ -1,2 +1,21 @@
 groups:
   - name: topolvm
+    rules:
+      - alert: AvailableBytesHighUtilizationInShortTerm
+        expr: |
+          sum(predict_linear(topolvm_volumegroup_available_bytes[1h], 3600*24)) without (instance,node) <= 0
+        labels:
+          severity: critical
+        for: 10m
+        annotations:
+          summary: Total available bytes of CS nodes will be exhausted in 1 day.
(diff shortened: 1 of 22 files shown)


monitoring/base/prometheus/alert_rules/topolvm.yaml
```diff
@@ -1,2 +1,21 @@
 groups:
   - name: topolvm
+    rules:
+      - alert: AvailableBytesHighUtilizationInShortTerm
+        expr: |
+          sum(predict_linear(topolvm_volumegroup_available_bytes[1h], 3600*24)) without (instance,node) <= 0
+        labels:
+          severity: critical
```
ymmt2005: These predictions should be considered as `warning`, I think.

dulltz: Agree it 🙆‍♂

PR #395, opened 2020-02-06
Rename maneki-apps project to tenant-apps-of-apps project

dulltz:
Rename maneki-apps to tenant-apps-of-apps because maneki-apps is confusing.


team-management/base/maneki/project.yaml
```diff
@@ -36,7 +36,7 @@ spec:
 apiVersion: argoproj.io/v1alpha1
 kind: AppProject
 metadata:
-  name: maneki-apps
+  name: tenant-app-of-apps
```
morimoto-cybozu: This project should be placed in "common" directory.

PR #396, opened 2020-02-06
ci: fix to copy files even when CI failed.

morimoto-cybozu:


morimoto-cybozu:
@ymmt2005 
That's a great idea!  I'll dedupe.

PR #481, opened 2020-04-03
grafana: remove $port variable from node-exporter-full

dulltz:


dulltz:
Close, continue to https://github.com/cybozu-go/neco-apps/pull/483

PR #490, opened 2020-04-07
[2020-04] Update the latter half of apps

tapih:
This is the latter half of the container image update on April 2020.
The former half is #491.


llamerada-jp:
This PR is incorporated into https://github.com/cybozu-go/neco-apps/pull/491

PR #531, opened 2020-04-30
grafana-operator: replace Role with ClusterRole to manage grafana resources in all namespaces

yokaze:


yokaze:
This PR is no longer needed, because we found `grafana-operator` does to allow to make Grafana instances in multiple namespaces.

PR #578, opened 2020-06-03
[CI] Stage stage-2020.06.03-9979

cybozu-neco:


kfyharukz:
Because of duplicate PR.

PR #589, opened 2020-06-05
ingress: add dummy HTTPProxies

dulltz:
This PR is to prevent envoy crush loop.
See: https://github.com/cybozu-go/neco/issues/1002

### DIFF
--- a/ingress/base/bastion/httpproxy.yaml
+++ b/ingress/base/bastion/httpproxy.yaml
@@ -1,0 +1,8 @@
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: dummy
+  namespace: ingress
+  annotations:
+    contour-plus.cybozu.com/exclude: "true"
+    kubernetes.io/ingress.class: bastion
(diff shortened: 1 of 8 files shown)


ingress/base/template/httpproxy.yaml
```diff
@@ -0,0 +1,57 @@
...
+spec:
+  virtualhost:
+    fqdn: dummy.ingress.dummy
+    tls:
+      secretName: dummy-httpproxy-tls
+  routes:
+    - conditions:
+        - prefix: /
+      services:
+        - name: dummy-tls
```
ysksuzuki, line 18: This service name should be `dummy` not `dummy-tls`, shouldn't it?

ingress/base/bastion/httpproxy.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: dummy
+  namespace: ingress
```
zoetrope, line 5: Is the namespace correct?
`ingress` namespace doesn't exist.
It should be `ingress-bastion`.

zoetrope: Sorry, I misunderstood.
The namespace will be overwrite in kustomization.yaml.

ingress/base/template/httpproxy.yaml
kfyharukz: You're right. I corrected it!

PR #593, opened 2020-06-08
Periodical update of containers & manifests

yokaze:
This PR for the periodical update of containers & manifests.
https://github.com/cybozu-private/neco-ops/issues/405

### DIFF
--- a/argocd/base/deployment.yaml
+++ b/argocd/base/deployment.yaml
@@ -6,7 +6,7 @@ spec:
   template:
     spec:
       containers:
-        - image: quay.io/cybozu/argocd:1.5.6.1
+        - image: quay.io/cybozu/argocd:1.5.6.2
           name: argocd-application-controller
 ---
 apiVersion: apps/v1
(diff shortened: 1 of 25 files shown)


argocd/base/deployment.yaml
```diff
@@ -6,7 +6,7 @@ spec:
   template:
     spec:
       containers:
-        - image: quay.io/cybozu/argocd:1.5.6.1
```
zoetrope, line 9: It's annoying to change the same tag name in multiple places.
How about using kustomize's imageTags?

test/Makefile
```diff
@@ -24,7 +24,7 @@ PROMTOOL_VERSION := 2.12.0
 TELEPORT_VERSION := 4.2.8
```
zoetrope, line 24: Doesn't this require an upgrade to 4.2.10?

argocd/base/deployment.yaml
kfyharukz: I agree.

test/Makefile
kfyharukz: Oh, I'm missing an update.

PR #616, opened 2020-06-19
Ingress watcher

dulltz:
Deploy the container created in this PR.
https://github.com/cybozu/neco-containers/pull/386


tapih:
This PR is not required any more, because ingress-watcher-global is deployed via cybozu-go/neco.
Close in favor of https://github.com/cybozu-go/neco-apps/pull/624 .

PR #620, opened 2020-06-23
rook: set tuneSlowDeviceClass to ceph-hdd

masa213f:


rook/overlays/gcp/ceph-hdd/cluster.yaml
```diff
@@ -15,6 +15,7 @@ spec:
     storageClassDeviceSets:
       - name: set1
         count: 3
+        tuneSlowDeviceClass: true
```
ymmt2005, line 18: @masa213f 
Why do you add this field?  It's the same as `base`.

ymmt2005: Never mind.  It's because `CephCluster` is a custom resource
and Kustomize cannot merge this field wisely.

masa213f: Yes. Unfortunately, Kustomize just replaces the entry of `storageClassDeviceSets`, cannot merge.

PR #629, opened 2020-06-30
unbound: add new application for unbound dns service

llamerada-jp:
Expose unbound to bastion.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -38,7 +38,6 @@ test-alert-rules:
 
 kustomize-check:
 	test "$$(ls ../argocd-config/base)" = "$$(ls ../argocd-config/overlays/stage0)"
-	test "$$(kustomize build ../argocd-config/overlays/stage0/ | grep "targetRevision:" | grep -v "stage" | wc -l)" = 0
 	./kustomize-check.sh
 
 code-check:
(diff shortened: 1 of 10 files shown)


test/Makefile
```diff
@@ -37,6 +37,8 @@ test-alert-rules:
 	promtool test rules alert_test/*.yaml
 
 kustomize-check:
+	test "$$(ls ../argocd-config/base)" = "$$(ls ../argocd-config/overlays/stage0)"
+	test "$$(kustomize build ../argocd-config/overlays/stage0/ | grep "targetRevision:" | grep -v "stage" | wc -l)" = 0
```
ymmt2005, line 41: This has been tested already at https://github.com/cybozu-go/neco-apps/blob/3123396d681a77f263b7444d062587c161a87c2a/test/validation_test.go#L55

test/suite_test.go
```diff
@@ -75,6 +75,7 @@ var _ = Describe("Test applications", func() {
 	Context("alertmanager", testAlertmanager)
 	Context("pushgateway", testPushgateway)
 	Context("ingress-health", testIngressHealth)
+	Context("unbound", testUnboundService)
```
ymmt2005, line 78: I don't think this test is worth existing.
Unbound itself has been tested, MetalLB has been tested, and there would be virtually no risk for degrading this added Service.

PR #672, opened 2020-07-16
teleport: Use ceph-ssd-block for storage class

tapih:
- Update teleport StratefulSet's storageClass to use ceph-ssd-block
- This requires some manual operations to sync the teleport application successfully
- Now teleport requires rook to be deployed beforehand, so change sync-waves for some applications.
  - The values of sync-waves were managed under `docs` as a markdown, but it's now managed under `test` to test sync-waves are set properly of not

Signed-off-by: Hiroshi Muraoka <h.muraoka714@gmail.com>


ymmt2005:
Please rebase it.

PR #706, opened 2020-07-27
test: use wget to download teleport binaries

ymmt2005:


test/Makefile
```diff
@@ -71,6 +71,7 @@ dctest-upgrade:
 	UPGRADE=1 ./test.sh
 
 setup:
+	apt-get update && apt-get -y install wget
```
tapih, line 74: Is "$(SUDO)" needed in this line?

ymmt2005, line 74:
```suggestion
	$(SUDO) apt-get update && $(SUDO) apt-get -y install wget
```

ymmt2005: Yes, fixed.

PR #718, opened 2020-07-29
clean DNS records not all at once

kfyharukz:
Delete DNS records for every fixed number, not all at once.

Signed-off-by: Kazuhito MATSUDA <kazuhito.matuda@gmail.com>

### DIFF
--- a/bin/clean-dns.sh
+++ b/bin/clean-dns.sh
@@ -1,4 +1,4 @@
-#! /bin/bash -e
+#! /bin/sh -e
 PROJECT=neco-test
 ZONE=gcp0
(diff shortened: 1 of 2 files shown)


bin/clean-dns.sh
```diff
@@ -0,0 +1,22 @@
+#! /bin/bash -e
```
ymmt2005, line 1: Please avoid bash scripts.  I don't understand bash specific extensions.
This is our SRE team's policy, too.

umezawatakeshi: This script is actually written in the range of bourne shells. I will rewrite shebang line.

PR #762, opened 2020-10-07
Stop deploying maneki-apps for dctest

zoetrope:

PR #765, opened 2020-10-08
[Rook/Ceph] stabilize neco-apps CI

llamerada-jp:
- Change Teleport's storage to TopoLVM instead of RBD.
- Move Rook to last of syncWeve.
- Ignore Rook on eventually loop to wait for sync of ArgoCD.
- Move test for Rook to Ceph test job.

### DIFF
--- a/argocd-config/base/maneki-apps.yaml
+++ b/argocd-config/base/maneki-apps.yaml
@@ -4,7 +4,7 @@ metadata:
   name: maneki-apps
   namespace: argocd
   annotations:
-    argocd.argoproj.io/sync-wave: "11"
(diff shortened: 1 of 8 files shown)


argocd-config/base/maneki-apps.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: maneki-apps
   namespace: argocd
   annotations:
-    argocd.argoproj.io/sync-wave: "11"
```
zoetrope, line 7: I'm concerned about conflicting with #762 .
But don't worry, I'll fix it.

zoetrope (approved):
LGTM

PR #787, opened 2020-10-19
rook: remove rook app from gcp CI to stabilize CI

llamerada-jp:


llamerada-jp:
operation missed

PR #789, opened 2020-10-20
rook: bump rook version to 1.4.6

llamerada-jp:

### DIFF
--- a/test/storage_test.go
+++ b/test/storage_test.go
@@ -222,8 +222,8 @@ func testDeploymentsVersion() {
 				}
 
 				for _, pod := range pods.Items {
-					if pod.Status.Phase != corev1.PodRunning {
-						fmt.Fprintf(GinkgoWriter, "pod status is not Runnning: ns=%s name=%s time=%s", pod.Namespace, pod.Name, time.Now())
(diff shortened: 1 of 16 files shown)


toshipp:
@kfyharukz @satoru-takeuchi 
All tests passed, would you review this?

test/storage_test.go
```diff
@@ -188,6 +190,86 @@ func testRookOperator() {
 	}
 }
 
+func testDeploymentsVersion() {
+	nss := []string{"ceph-hdd", "ceph-ssd"}
+	for _, ns := range nss {
+		It("should rook's Deployments version of "+ns+" is equal to operators version", func() {
```
kfyharukz, line 196: I think you'd better use `By` sentences to show what you intend the codes.

kfyharukz, line 196: I think this test name isn't appropriate because this test includes a health check of the `Deployments` and the ceph cluster.  
And, you'd better use `By` sentences to show what you intend with the codes.

kfyharukz (changes requested):

kfyharukz (approved):
LGTM

PR #806, opened 2020-11-06
team-management: configure rbac to create obc resource for tenant users

llamerada-jp:


ymmt2005 (approved):
LGTM

PR #814, opened 2020-11-11
Remove destination.namespace from Application.

kfyharukz:
Signed-off-by: Kazuhito MATSUDA <kazuhito.matuda@gmail.com>


kfyharukz:
Close because this is not the main cause of too long sync duration.

PR #783, opened 2020-10-16
202009: Regular container update

ysksuzuki:
Regular container update.
Signed-off-by: Yusuke Suzuki <yusuke-suzuki@cybozu.co.jp>

PR #816, opened 2020-11-12
Check application status in syncWave order 

tapih:


masa213f:
The content of this PR is included in https://github.com/cybozu-go/neco-apps/pull/783 .

PR #825, opened 2020-11-16
remove useless prometheus relabel configs.

umezawatakeshi:
prometheus.yaml have some `action: drop` relabel configs to drop some metrics. However, those configs does not work and the metrics have been already used. So, remove the useless configs.


morimoto-cybozu (approved):
LGTM

PR #836, opened 2020-11-19
Add teleport proxy test which confirms that teleport client can access kube-apiserver

tapih:
Update of teleport from v4.1 to v4.2 caused a degradation that the URL of kube-apiserver is given like 0.0.0.0:3206.
This PR contains a new test case to prevent this behavior from occurring again.


masa213f (approved):
LGTM

PR #884, opened 2020-12-04
coil: Enable sessionAffinity for egresses

tapih:
I investigated https://github.com/torvalds/linux/blob/bbe2ba04c5a92a49db8a42c850a5a2f6481e47eb/net/netfilter/ipvs/ip_vs_core.c
and found that the session affinity timeout for UDP is updated every time packets are sent or received.
This means that the source hashing option is not needed any more and we should just use `sessionAffinity` to keep long-lived sessions  alive.

### DIFF
--- a/coil/base/egress.yaml
+++ b/coil/base/egress.yaml
@@ -7,7 +7,10 @@ spec:
   replicas: 2
   destinations:
   - 0.0.0.0/0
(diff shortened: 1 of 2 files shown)


ymmt2005 (approved):
LGTM

PR #885, opened 2020-12-04
mackerel-agent: store stateful data in Ceph RBD volume

ymmt2005:
`mackerel-agent` stores its host ID under `/var/lib/mackerel-agent` directory.
To keep the host ID between Pod recreation, the volume for the directory should be persistent.


yokaze (approved):
LGTM

PR #892, opened 2020-12-07
[rook/ceph] set request parameters for the pods

llamerada-jp:

### DIFF
--- a/rook/base/ceph-hdd/cluster.yaml
+++ b/rook/base/ceph-hdd/cluster.yaml
@@ -7,6 +7,37 @@ metadata:
     argocd.argoproj.io/sync-wave: "1"
 spec:
   dataDirHostPath: /var/lib/rook
+  resources:
+    mgr:
+      requests:
+        cpu: "500m"
+        memory: "512Mi"
+    mon:
+      requests:
+        cpu: "500m"
+        memory: "1Gi"
+    ## send bug report https://github.com/rook/rook/issues/6789
+    # osd:
+    #   requests:
+    #     cpu: "500m"
+    #     memory: "2Gi"
+    prepareosd:
+      requests:
+        cpu: "500m"
+        memory: "50Mi"
+    crashcollector:
+      requests:
+        cpu: "500m"
+        memory: "60Mi"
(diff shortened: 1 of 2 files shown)


rook/base/ceph-hdd/cluster.yaml
```diff
@@ -37,11 +68,12 @@ spec:
     storageClassDeviceSets:
       - name: set1
         count: 18
-        tuneDeviceClass: true
+        ## send bug report https://github.com/rook/rook/issues/6789
         resources:
           requests:
             cpu: "500m"
-            memory: "4Gi"
```
binoue, line 44: Why did you reduce the request memory?
Based on [the Ceph document](https://docs.ceph.com/en/latest/start/hardware-recommendations/#memory), osd may result in degraded performance with between 2GB and 4GB memory.

rook/base/ceph-ssd/cluster.yaml
```diff
@@ -7,6 +7,37 @@ metadata:
...
   dataDirHostPath: /var/lib/rook
+  resources:
+    mgr:
+      requests:
+        cpu: "500m"
+        memory: "512Mi"
+    mon:
+      requests:
+        cpu: "500m"
+        memory: "1Gi"
```
ymmt2005, line 18: Does `mon` really consume memory more than 1GiB?

toshipp: I don't know that, but rook recommends it.
https://github.com/rook/rook/blob/master/Documentation/ceph-cluster-crd.md#cluster-wide-resources-configuration-settings

rook/base/ceph-hdd/cluster.yaml
toshipp: Yes, rook document says 2G is the minimum requirement.
If OSD requires more memory, this is a request, so OSD can consume more memory.
This is a baseline, we would tune the value later.

rook/base/ceph-ssd/cluster.yaml
ymmt2005: I see.

toshipp (approved):
LGTM

PR #900, opened 2020-12-08
monitoring: fix alert rules to use neco reboot-worker

yokaze:
This PR adds and fixes alert rules for `neco reboot-worker`.

### What this PR does: 

- `CKEOperationTakesLongTime`
  Reboot phase is removed from alert calculation.
- `CKESabakanIntegrationDoesNotPerformAnyOps`
  It is silenced when reboot operation is ongoing.
- `CKERebootQueueStuck`
  It is added to alert reboot queue being stuck for 60mins.
- `K8sAPIServersDegraded`
  Extend from 10mins to 15mins.
- `KubeDaemonSetRolloutStuck`
  Extend from 15mins to 30mins.
- `KubeNodeNotReady`
  It is silenced when the node is rebooted.
- `KubeNodeUnschedulable`
  Extend from 15mins to 60mins.

### What this PR is not for:

- `KubePodCrashLooping`
  It is already treated in #891.
- `KubeDeploymentReplicasMismatch`
- `KubePodNotReady`
  They are not treated in this PR.

### DIFF
--- a/monitoring/base/prometheus/alert_rules/cke.yaml
+++ b/monitoring/base/prometheus/alert_rules/cke.yaml
@@ -21,7 +21,7 @@ groups:
           severity: minor
       - alert: CKEOperationTakesLongTime
         expr: |
-          sum(cke_operation_phase{phase!="completed"}) > 0
+          sum(cke_operation_phase{phase!~"completed|reboot-nodes"}) > 0
         labels:
           severity: warning
         for: 30m
@@ -39,12 +39,21 @@ groups:
           runbook: TBD
       - alert: CKEDoesNotPerformAnyOps
         expr: |
-          rate(cke_operation_phase_timestamp_seconds[1h]) == 0
+          rate(cke_operation_phase_timestamp_seconds[5m]) == 0
         labels:
           severity: warning
-        for: 15m
+        for: 1h
         annotations:
           summary: CKE does not perform any operations for more than 1 hour.
+          runbook: TBD
+      - alert: CKERebootQueueStuck
+        expr: |
+          sum(rate(cke_reboot_queue_entries[5m])) == 0 and sum(cke_reboot_queue_entries) > 0
+        labels:
+          severity: warning
+        for: 1h
+        annotations:
+          summary: CKE reboot queue is stuck for more than 1 hour.
           runbook: TBD
       - alert: CKESabakanIntegrationSeemsToBeFailed
         expr: |
@@ -57,7 +66,7 @@ groups:
           runbook: TBD
       - alert: CKESabakanIntegrationDoesNotPerformAnyOps
         expr: |
-          rate(cke_sabakan_integration_timestamp_seconds[1h]) == 0
+          sum(rate(cke_sabakan_integration_timestamp_seconds[5m])) == 0 and sum(cke_operation_phase{phase="reboot-nodes"}) == 0
         labels:
           severity: warning
         for: 60m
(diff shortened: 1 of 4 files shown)


monitoring/base/prometheus/alert_rules/cke.yaml
```diff
@@ -57,7 +66,7 @@ groups:
           runbook: TBD
       - alert: CKESabakanIntegrationDoesNotPerformAnyOps
         expr: |
-          rate(cke_sabakan_integration_timestamp_seconds[1h]) == 0
+          sum(rate(cke_sabakan_integration_timestamp_seconds[5m])) == 0 and sum(cke_operation_phase{phase="reboot-nodes"}) == 0
```
masa213f, line 69: How about this?
```
((cke_sabakan_integration_timestamp_seconds - cke_sabakan_integration_timestamp_seconds offset 5m) == 0)
  and ignoring(phase)
(cke_operation_phase{phase="reboot-nodes"} == 0)
```
or
```
((cke_sabakan_integration_timestamp_seconds - cke_sabakan_integration_timestamp_seconds offset 5m) == 0)
  unless ignoring(phase)
(cke_operation_phase{phase="reboot-nodes"} == 1)
```
The result has `instance` label. We can improve the messages which will be posted in slack.

monitoring/base/prometheus/alert_rules/kubernetes.yaml
```diff
@@ -428,17 +428,17 @@ groups:
           summary: '{{ $labels.node }} has been unready for more than 15 minutes.'
           runbook: TBD
         expr: |
-          kube_node_status_condition{job="kube-state-metrics",condition="Ready",status="true"} == 0
+          sum by(node) (kube_node_status_condition{job="kube-state-metrics",condition="Ready",status="true"}) == 0 and sum by(node) (kube_node_spec_unschedulable{job="kube-state-metrics"}) == 0
```
masa213f, line 431: How about this?
```
(kube_node_status_condition{job="kube-state-metrics",condition="Ready",status="true"} == 0)
  and ignoring(condition, status)
(kube_node_spec_unschedulable{job="kube-state-metrics"} == 0)
```

or 
```
(kube_node_status_condition{job="kube-state-metrics",condition="Ready",status="true"} == 0)
  unless ignoring(condition, status)
(kube_node_spec_unschedulable{job="kube-state-metrics"} == 1)
```

masa213f (changes requested):

monitoring/base/prometheus/alert_rules/cke.yaml
yokaze: When we use `ignoring` with two different metrics, we need to know the complete set of labels in advance. It seems a bit difficult.
Phrases like `sum by` can keep necessary labels without knowing them exactly, hence it may improve readability.
How do you think?

masa213f: > It seems a bit difficult.

We need to know the labels to write a proper rule, I think. Is it difficult?

- `cke_sabakan_integration_timestamp_seconds` has only 2 labels. (`job` and `instance` labels)
- `cke_operation_phase` has only 3 labels. (`job`, `instance` and ` phase` labels)

With Grafana it's easy to see.

Also, I don't understand the intent of summarizing metrics by sum or max.

yokaze: ic, I'll fix the expression as you wrote. Thanks for the comment.

yokaze:
This PR should be merged with squash.

masa213f (approved):

PR #903, opened 2020-12-09
neco-admission: Add preStop hook for graceful termination

tapih:
Implements https://github.com/cybozu-go/neco/issues/1255

### DIFF
--- a/neco-admission/base/deployment.yaml
+++ b/neco-admission/base/deployment.yaml
@@ -56,7 +56,11 @@ spec:
             path: /metrics
             port: 8080
             scheme: HTTP
-      terminationGracePeriodSeconds: 10
+        lifecycle:
+            preStop:
+              exec:
+                command: ["sh", "-c", "sleep 5"]
+      terminationGracePeriodSeconds: 15
       volumes:
         - name: certs
           secret:
--- a/neco-admission/base/deployment.yaml
+++ b/neco-admission/base/deployment.yaml
@@ -34,7 +34,7 @@ spec:
         args:
         - --cert-dir=/certs
         - --httpproxy-default-class=forest
-        image: quay.io/cybozu/neco-admission:0.6.1
+        image: quay.io/cybozu/neco-admission:0.6.2
         resources:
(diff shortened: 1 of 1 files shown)


neco-admission/base/deployment.yaml
```diff
@@ -56,7 +56,11 @@ spec:
             path: /metrics
             port: 8080
             scheme: HTTP
-      terminationGracePeriodSeconds: 10
+        lifecycle:
+            preStop:
+              exec:
+                command: ["sh", "-c", "sleep 5"]
+      terminationGracePeriodSeconds: 15
```
ymmt2005, line 63: I don't think there is any need to extend `terminationGracePeriodSeconds`.

Besides, the container image of `neco-admission` is based on `scratch`, so
it has no `sh` nor `sleep`.  I'm also curious about the reason for `sh -c`.
Is `command: ["sleep", "5"] not enough?

tapih: > I don't think there is any need to extend `terminationGracePeriodSeconds`.

I think the grace termination period starts at the same time the preStop hook does.
You mean 15 seconds for terminationGracePeriodSeconds is just too long for this container, right?

> Besides, the container image of `neco-admission` is based on `scratch`, so it has neither `sh` nor `sleep`.

You are absolutely correct. I'm gonna replace the base image with ubuntu.

> I'm also curious about the reason for `sh -c`. Is command: `["sleep", "5"]` not enough?

I followed the way the contour manifest and the squid manifest do and I'm guessing that they used to include multiline shell scripts like `sleep 5; ls; sleep 5`, which is easy to read with `sh -c`.
I didn't mind it very much but as you said, just `["sleep", "5"]` is enough in this case.

ymmt2005: > You mean 15 seconds for terminationGracePeriodSeconds is just too long for this container, right?

Yes.  The webhook can process the request really quickly.

ymmt2005 (approved):
LGTM

PR #923, opened 2020-12-11
[CI] Release release-2020.12.11-15112

cybozu-neco:


tapih:
partly duplicated with https://github.com/cybozu-go/neco-apps/pull/929, so close

PR #926, opened 2020-12-11
grafana-operator: Upgrade to 3.7.0

tapih:
I confirmed that there is only the difference that I recognize should be according to the release note by executing `kustomize build .` before and after the update.

Ref https://github.com/cybozu/neco-containers/pull/475


kfyharukz (dismissed):
LGTM

binoue (approved):
LGTM

PR #941, opened 2020-12-15
add monitoring configs for VictoriaMetrics itself

umezawatakeshi:

### DIFF
--- a/monitoring/base/victoriametrics/rules/monitoring-scrape.yaml
+++ b/monitoring/base/victoriametrics/rules/monitoring-scrape.yaml
@@ -1,0 +1,109 @@
+apiVersion: operator.victoriametrics.com/v1beta1
+kind: VMPodScrape
+metadata:
+  name: victoriametrics-operator
+  namespace: monitoring
+  labels:
+    smallset: "true"
+spec:
+  namespaceSelector:
+    matchNames: [monitoring]
+  selector:
+    matchLabels:
+      control-plane: victoriametrics-operator
+  podMetricsEndpoints:
+  - port: http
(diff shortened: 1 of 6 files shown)


monitoring/base/victoriametrics/rules/monitoring-scrape.yaml
```diff
@@ -0,0 +1,134 @@
...
+  labels:
+    smallset: "true"
+spec:
+  namespaceSelector:
+    matchNames: [monitoring]
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: vmalert
+      managed-by: vm-operator
+      vmalert-smallset-cluster: "true"
```
masa213f, line 96: I'm a little confused. If we use two sets of vmalert(`vmalert-smallset-0`, `vmalert-smallset-1`) and they work independently, it is better to use the following labels.

```suggestion
      app.kubernetes.io/instance: vmalert-smallset-0
      app.kubernetes.io/name: vmalert
      managed-by: vm-operator
```

monitoring/base/victoriametrics/rules/monitoring-scrape.yaml
```diff
@@ -0,0 +1,134 @@
...
+  podMetricsEndpoints:
+  - relabelConfigs:
+      - replacement: vmalert-smallset
```
masa213f, line 99: Similar to the above comment, it is better to distinguish between `vmalert-smallset-0` and `vmalert-smallset-1`, I think.

masa213f (dismissed):
LGTM

masa213f (approved):

PR #945, opened 2020-12-16
[CI] Stage stage-2020.12.16-15336

cybozu-neco:


tapih:
Close due to duplication with https://github.com/cybozu-go/neco-apps/pull/950

PR #948, opened 2020-12-16
customize envoy log fields

binoue:
Custom JSON fields for Envoy access logs is enabled since Contour v1.10.0.
By using this function, we can have many additional fields like `response_code_details`, `response_code_details, connection_time_details` in our Envoy access logs.

Therefore, in this PR, I added the following log fields in addition to the default fields.
- `response_code_details`
- `connection_time_details` 
- `request_duration` 
- `response_duration` 
- `upstream_transport_failure_reason`


tapih (approved):
LGTM

PR #954, opened 2020-12-18
cert-manager: Use ZeroSSL for stage0.

kfyharukz:
This PR replaces the ClusterIssuer (name: clouddns) from Let's Encrypt to ZeroSSL.
To enable the ClusterIssuer, we need the operation creating Secret that includes EAB HMAC Key.


ymmt2005 (approved):
LGTM

PR #958, opened 2020-12-21
Revert "argocd: Upgrade dex to v2.27.0"

binoue:
After upgrading dex from v2.22.0 to v2.27.0, argocd login via cli does not work,
Therefore, this PR revert the dex upgrade.


tapih (approved):
LGTM

PR #963, opened 2020-12-21
[CI] Stage stage-2020.12.21-15470

cybozu-neco:


tapih:
In favor of https://github.com/cybozu-go/neco-apps/pull/967

PR #982, opened 2020-12-23
use VMNodeScrape and revise existing scraping configuration

umezawatakeshi:
- upgrade VictoriaMetrics operator to v0.7.1
- use VMNodeScrape instead of additionalScrapeConfigs
- use `spec.discoveryRole: service` in VMServiceScrape for kube-state-metrics job instead of mimicking `role: service` using `role: endpoints`
- do not remove port number part in `instance` target label
 

masa213f (approved):
LGTM

PR #1024, opened 2021-01-08
Add boot server alerts.

kfyharukz:
This PR adds some alerts for boot servers including for detecting CKE failure.


binoue (approved):
LGTM

kfyharukz:
These alerts will be inserted into the VictoriaMetrics large cluster.
(The VMSingle should be a minimum set for monitoring the components to run the large cluster

PR #1028, opened 2021-01-08
add PVC to vmalertmanager

umezawatakeshi:
Add PVC to vmalertmanager.
STS should be deleted manually after this PR is applied to stage/prod because STSs cannot be modified except  'replicas', 'template', and 'updateStrategy'.


yokaze (approved):
LGTM

zoetrope (approved):

PR #1035, opened 2021-01-12
make vmagent/vmalert see CRs in monitoring NS only

umezawatakeshi:
To reduce load of API server, see only monitoring NS.
If namespace selector is omitted (nil), they see only the NS which they are in.


zoetrope (approved):
LGTM

PR #1042, opened 2021-01-13
Remove unnecessary webhook ignore labels

kfyharukz:
This PR removes unnecessary webhook ignore labels.
Note that `vnetworkpolicy` rejects less than `admission.cybozu.com/min-policy-order` annotation value, not including "equal" value (default is `1000.0`).
See: https://github.com/cybozu/neco-containers/pull/497


morimoto-cybozu (approved):
LGTM

Issue #1025, opened 2021-01-08
Update manifests for Kubernetes 1.19


ymmt2005:
## What

Manifests must work for both Kubernetes 1.18 and 1.19 before upgrading Kubernetes.
Specifically, MetalLB has a stale label `beta.kubernetes.io/os` and therefore must be updated.

## Technical information

Deprecated API versions:

- apiextensions.k8s.io/v1beta1 is deprecated in favor of apiextensions.k8s.io/v1
- apiregistration.k8s.io/v1beta1 is deprecated in favor of apiregistration.k8s.io/v1
- authentication.k8s.io/v1beta1 and authorization.k8s.io/v1beta1 are deprecated in 1.19 in favor of v1 levels and will be removed in 1.22
- autoscaling/v2beta1 is deprecated in favor of autoscaling/v2beta2
- coordination.k8s.io/v1beta1 is deprecated in 1.19, targeted for removal in 1.22, use v1 instead.
- storage.k8s.io/v1beta1 is deprecated in favor of storage.k8s.io/v1

Renamed metrics:

- Kubelet: following metrics have been renamed: kubelet_running_container_count --> kubelet_running_containers kubelet_running_pod_count --> kubelet_running_pods
- Rename pod_preemption_metrics to preemption_metrics.
- Kube-scheduler: The metric name scheduler_total_preemption_attempts has been renamed to scheduler_preemption_attempts_total.

## How

- Update MetalLB
- Update metrics names

## Checklist

- [x] Finish implentation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1066, opened 2021-01-18
Update manifests for Kubernetes 1.19

ymmt2005:
#1025

Although `apiextensions.k8s.io/v1beta1` has been deprecated, there are many CRDs in upstream
manifests that remain `apiextensions.k8s.io/v1beta1`.  We cannot convert them at this point.

Metrics will be updated separately in #1067.
TopoLVM will be updated separately in #1070.


ysksuzuki (approved):
LGTM

PR #1068, opened 2021-01-19
fix scraping CR for VM smallset

umezawatakeshi:
- use VMServiceScrape instead of VMPodScrape for vmagent/vmalert
- add `app.k8s.io/component` label


yokaze (approved):
LGTM

PR #1084, opened 2021-01-19
do not make VictoriaMetrics operator create self VMServiceScrape

umezawatakeshi:
- we don't use VMServiceScrape which is created by the operator for VMAgent, VMAlert etc.
- we never require Prometheus operator CR conversion

see also: https://github.com/VictoriaMetrics/operator/blob/v0.7.1/vars.MD


yokaze (approved):
LGTM

PR #1087, opened 2021-01-20
remove topolvm's alert

llamerada-jp:
This is because the application team should monitor the capacity of each persistent volume,
rather than the admin monitoring the free space of each individual node.


toshipp (approved):
LGTM

PR #1105, opened 2021-01-22
test: remove "cluster.local" domain suffix when accessing from nodes.

morimoto-cybozu:


binoue (approved):
LGTM

PR #1108, opened 2021-01-22
rook alert is not enabled in VictoriaMetrics

umezawatakeshi:
fix #1044


toshipp (approved):
LGTM

PR #1117, opened 2021-01-25
Changed metrics for detecting registry values

zoetrope:
`registry_storage_cache_total` is not a metric that is always exposed.
Instead, use `registry_storage_action_seconds_bucket`.


yokaze:
@zoetrope 
Did you check the replacement works fine on the Grafana dashboard?
Since the PR uses bucket metrics instead of a scalar metrics, I suspect it needs additional work to do.

zoetrope:
@yokaze 
Yes, I checked it. It works.

yokaze (approved):
LGTM

PR #1122, opened 2021-01-26
Update moco to v0.4.0

masa213f:


yokaze (dismissed):
LGTM

test/moco_test.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+func prepareMoco() {
+	It("should deploy mysqlcluster", func() {
+		// This manifest is based on the this example (https://github.com/cybozu-go/moco/blob/v0.4.0/docs/example_mysql_cluster.md).
+		// Changed as follows.
+		// - Change the namespace of all resources.
+		// - Change the tag of quay.io/cybozu/moco-mysql image.
+		// - Remove the `.spec.serviceTemplate` field from MySQLCluster resource.
+		manifest := `apiVersion: moco.cybozu.com/v1alpha1
```
zoetrope, line 21: Could you extract this manifest as a single file?
I think a long yaml embedded in go is difficult to maintain.

ref: https://github.com/cybozu-go/coil/blob/44aa1e256e8ead380cf1a36d2fa413747db0c50f/v2/e2e/controller_test.go#L20

test/moco_test.go
```diff
@@ -0,0 +1,216 @@
...
+metadata:
+  name: my-cluster
+  namespace: test-moco
+spec:
+  replicas: 3
+  podTemplate:
+    spec:
+      containers:
+      - name: mysqld
+        image: quay.io/cybozu/moco-mysql:8.0.20.9
```
zoetrope, line 32: Isn't 8.0.18 is suitable?
Because that is what user is mainly using.

test/moco_test.go
```diff
@@ -0,0 +1,216 @@
...
+				return err
+			}
+
+			if string(stdout) != "True" {
+				return errors.New("MySQLCluster is not ready")
+			}
+			return nil
+		}).Should(Succeed())
+
+		By("copying kubectl-moco")
```
zoetrope, line 201: Please add the following todo comment.
```
//TODO: kubectl-moco will be included in neco.deb in the future, so remove the following codes
```

test/moco_test.go
masa213f: That's a good idea.
But this manifest will become shorter with the next version of MOCO.
Because the container definitions and ConfigMaps for the FileBeat can be omitted.

Then this manifest will be short as same as any other test manifests.
So I don't think need to divide it as a file now.

zoetrope: I see.

zoetrope (dismissed):
LGTM

zoetrope (approved):

PR #1144, opened 2021-02-03
fix NodeExporterDown alert on Prometheus

umezawatakeshi:
`job="node-exporter"` metrics have port number in `instance` label (i.e. `instance="10.0.0.1:9100"`, not `instance="10.0.0.1"`), even on Prometheus.


morimoto-cybozu (approved):
LGTM

PR #1153, opened 2021-02-04
monitoring: add Kubernetes labels to node/namespace metrics

ymmt2005:
Starting with kube-state-metrics v2.0.0, labels metrics such as
`kube_namespace_labels` or `kube_node_labels` does not include
Kubernetes resource labels.

https://github.com/kubernetes/kube-state-metrics/issues/1270

Since we are using some Kubernetes labels, this commit restores them.


zoetrope (approved):
LGTM

PR #1179, opened 2021-02-10
logging: add Eventually for checkLog method to stablize CI

llamerada-jp:


toshipp (approved):
LGTM

PR #1212, opened 2021-02-18
argocd: add reference to garoon-apps repository

yokaze:


masa213f (approved):
LGTM

PR #1214, opened 2021-02-18
monitoring: fix alert rules for DaemonSets

ymmt2005:
Unreachable nodes can be tainted with "NoSchedule" effect
instead of "NoExecute".


morimoto-cybozu (approved):
paired

PR #1215, opened 2021-02-18
Apply placemat v2

ysksuzuki:
This PR modifies tests to work with both placemat versions 1 and 2.

### DIFF
--- a/test/teleport_test.go
+++ b/test/teleport_test.go
@@ -63,6 +63,9 @@ func teleportSSHConnectionTest() {
 		"--output=jsonpath={.status.loadBalancer.ingress[0].ip}")
 	Expect(err).ShouldNot(HaveOccurred(), "stderr=%s", stderr)
 	addr := string(stdout)
+	b, err := ioutil.ReadFile("/etc/hosts")
+	Expect(err).NotTo(HaveOccurred())
+	Expect(ioutil.WriteFile("./hosts", b, 0644)).NotTo(HaveOccurred())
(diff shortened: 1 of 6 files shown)


test/teleport_test.go
```diff
@@ -63,6 +63,9 @@ func teleportSSHConnectionTest() {
 		"--output=jsonpath={.status.loadBalancer.ingress[0].ip}")
 	Expect(err).ShouldNot(HaveOccurred(), "stderr=%s", stderr)
 	addr := string(stdout)
+	b, err := ioutil.ReadFile("/etc/hosts")
+	Expect(err).NotTo(HaveOccurred())
+	Expect(ioutil.WriteFile("./hosts", b, 0644)).NotTo(HaveOccurred())
 	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
ymmt2005, line 69: I cannot understand these parts...  Could you explain them?

ysksuzuki: This saves a backup of `/etc/hosts` as `./hosts` before adding `teleport.gcp0.dev-ne.co` entry, and restore it at the end of the tests instead of removing the entry using sed, since `/etc/host` cannot be modified with sed inside a container.

ymmt2005: So this rewrite the container host's `/etc/hosts`, right?
What happens when placemat2 runs on `stage0`?  We need to rewrite the Node's `/etc/hosts`?

ymmt2005 (approved):
LGTM!

PR #1220, opened 2021-02-18
loki: shorten termination period to avoid an alert when reboot cluster

llamerada-jp:


ymmt2005 (approved):
LGTM

ymmt2005:
This is insufficient because upgrading the stateful set takes 80 minutes...
will fix in #1223.

PR #1285, opened 2021-03-05
Add name label to app-ept-wiki namespace

zoetrope:
* Because I want to set NetworkPolicy using this label.


kfyharukz (approved):
LGTM

morimoto-cybozu:
@zoetrope 
Have you tried with the `projectcalico.org/name` label in `namespaceSelector`?
Though this label is not set explicitly, it may work as a "virtual" label.

zoetrope:
unnecessary

PR #1290, opened 2021-03-05
Add startupProbe for vmstorage

zoetrope:
* Because vmstorage sometimes takes a long time to start up.

### DIFF
--- a/monitoring/base/victoriametrics/vmcluster-largeset.yaml
+++ b/monitoring/base/victoriametrics/vmcluster-largeset.yaml
@@ -33,6 +33,14 @@ spec:
             app.kubernetes.io/instance: vmcluster-largeset
             app.kubernetes.io/name: vmstorage
             managed-by: vm-operator
+    containers:
+      - name: vmstorage
+        startupProbe:
+          httpGet:
+            path: /health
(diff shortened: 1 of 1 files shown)


monitoring/base/victoriametrics/vmcluster-largeset.yaml
```diff
@@ -33,6 +33,14 @@ spec:
             app.kubernetes.io/instance: vmcluster-largeset
             app.kubernetes.io/name: vmstorage
             managed-by: vm-operator
+    containers:
+      - name: vmstorage
+        startupProbe:
+          httpGet:
+            path: /
```
umezawatakeshi, line 40: I think it would be nice to use the same path as liveness probe.
```
$ kubectl get pod -n monitoring vmstorage-vmcluster-largeset-0 -o yaml
(snip)
    livenessProbe:
      failureThreshold: 3
      httpGet:
        path: /health
        port: 8482
        scheme: HTTP
(snip)
```

zoetrope: thanks!

umezawatakeshi (approved):

PR #1315, opened 2021-03-10
refine Grafana dashboards

umezawatakeshi:
- do not specify concrete instance value to query
- show all namespaces using templating
- use same timerange across dashboards and don't refresh by default
  - note: "Node Exporter Full" dashboard is not modified because its definition is not included in neco-apps but is downloaded from grafana.com


tapih (approved):
LGTM

PR #1323, opened 2021-03-15
Add a test to verify all permissions for tenants' CRs

binoue:
This PR adds a test to verify all permissions for tenants' CRs


masa213f (changes requested):
Could you sort resources?

test/team-management_test.go
```diff
@@ -224,6 +338,12 @@ func testTeamManagement() {
 					}
 				}
 
+				// check prohibited resources
```
masa213f, line 341: Please write the reason why expectedVerbs and actualVerbs are not updating here.

masa213f (approved):
LGTM

PR #1340, opened 2021-03-23
victoriametrics: pod of smallset run in the same node

llamerada-jp:


toshipp (approved):
LGTM

PR #1345, opened 2021-03-25
Remove Placemat v1 compatible codes

ysksuzuki:
This PR removes placemat v1 compatible codes. Also, reverts Teleport test to run on the test host. 
The original teleport test modifies `/etc/hosts` using sed and it leads to the error when the test runs inside a k8s container. So this PR fixes it to stop using sed as well.

### DIFF
--- a/test/teleport_test.go
+++ b/test/teleport_test.go
@@ -63,6 +63,10 @@ func teleportSSHConnectionTest() {
 		"--output=jsonpath={.status.loadBalancer.ingress[0].ip}")
 	Expect(err).ShouldNot(HaveOccurred(), "stderr=%s", stderr)
 	addr := string(stdout)
+	// Save a backup before editing /etc/hosts
+	b, err := ioutil.ReadFile("/etc/hosts")
+	Expect(err).NotTo(HaveOccurred())
+	Expect(ioutil.WriteFile("./hosts", b, 0644)).NotTo(HaveOccurred())
 	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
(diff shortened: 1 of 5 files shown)


test/teleport_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
```
tapih, line 11: `io/ioutil` is deprecated in Go [1.16](https://golang.org/doc/go1.16#ioutil). Please use `os` instead.

test/teleport_test.go
```diff
@@ -185,29 +180,21 @@ func teleportSSHConnectionTest() {
...
-
-	By("clearing teleport_cookie.txt")
-	_, stderr, err = ExecAt(boot1, "rm", filename)
-	Expect(err).ShouldNot(HaveOccurred(), "stderr=%s", stderr)
-
 	By("clearing /etc/hosts")
-	_, stderr, err = ExecAt(boot1, "sudo", "sed", "-i", "-e", "/teleport.gcp0.dev-ne.co/d", "/etc/hosts")
-	Expect(err).ShouldNot(HaveOccurred(), "stderr=%s", stderr)
+	// /etc/hosts cannot be modified via sed directly inside a container because that is a mount point
+	b, err = ioutil.ReadFile("./hosts")
```
tapih, line 194: This can be purged if we use `curl --resolve`.

tapih (changes requested):

test/teleport_test.go
ysksuzuki: I though I could, but unfortunatelly `tsh` seems to require `/etc/hosts` entry.

tapih (approved):
LGTM

PR #1357, opened 2021-03-30
loki: change loki deploy mode

toshipp:
- change loki deploy mode from monolithic to microservice.
  - use helm for consul because it can not be deployed by jsonnet provided from loki upstream.
- add loki-canary to monitor loki and scraping rules.


umezawatakeshi (approved):
LGTM

PR #1360, opened 2021-04-02
Increase maxScrapeSize

binoue:
As we faced the scraping error caused by the response size of the scraping metrics, this PR increased the maxScrapeSize parameter of VictoriaMetrics Agent.


morimoto-cybozu (approved):
LGTM

PR #1369, opened 2021-04-05
Remove ingress-bastion httpproxy

binoue:
This PR removes ingress-bastion httpproxy because ingress-watcher will be purged from neco


morimoto-cybozu (approved):
LGTM

PR #1370, opened 2021-04-05
monitoring: update VictoriaMetrics operator to 0.12.2

morimoto-cybozu:
This PR updates VictoriaMetrics operator from 0.10.0 to 0.12.2, and refines its manifests as follows.

* use embedded PodDisruptionBudget; introduced in 0.12.0
* remove `topologySpreadConstraints`s because they are duplicates of the default setting in Neco environment
  cf. https://github.com/cybozu-go/neco/blob/main/etc/cke-template.yml
* restore `livenessProbe` for vmstorage; removed in 0.12.1
* fix port name of vmalertmanager for scraping; changed in 0.11.0


umezawatakeshi (approved):
LGTM

PR #1374, opened 2021-04-06
teleport sync=NORMAL

masa213f:


masa213f:
After this configuration, dctest became flaky.
This configuration is too much load for dctest.

PR #1389, opened 2021-04-09
Fix test for reinstallation for cke-localproxy

binoue:
This PR lets neco-apps CI pass once after cke-localproxy is installed.


ymmt2005 (approved):
LGTM

PR #1391, opened 2021-04-09
test: fix reboot test for new "neco power"

morimoto-cybozu:
This PR fixes the reboot test for the new implementation of `neco power`.

* change the subcommand name from `ipmipower` to `power`
* fix not to call `neco power start` for the running boot servers
* remove the retries of `neco power stop/start`
* remove the special treatment of rack-3


ymmt2005 (approved):
LGTM

PR #1396, opened 2021-04-13
Remove temporal image validation

binoue:
This PR removes temporal image validation

### DIFF
--- a/neco-admission/base/deployment.yaml
+++ b/neco-admission/base/deployment.yaml
@@ -43,12 +43,8 @@ spec:
         - --valid-image-prefix=docker.elastic.co/
         - --valid-image-prefix=quay.io/gravitational/teleport-ent
         - --valid-image-prefix=gcr.io/datadoghq/agent
-        # TODO: remove this after maneki complete the transition to gcr.io
-        - --valid-image-prefix=datadog/agent,docker.io/datadog/agent
         # TODO: remove this after we build our own images for Rook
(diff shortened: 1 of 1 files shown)


neco-admission/base/deployment.yaml
```diff
@@ -45,10 +45,6 @@ spec:
         - --valid-image-prefix=gcr.io/datadoghq/agent
         # TODO: remove this after maneki complete the transition to gcr.io
         - --valid-image-prefix=datadog/agent,docker.io/datadog/agent
-        # TODO: remove this after we build our own images for Rook
-        - --valid-image-prefix=k8s.gcr.io/sig-storage/,quay.io/cephcsi/
```
ymmt2005, line 49: @llamerada-jp @toshipp @satoru-takeuchi 
Could you confirm this?

binoue: We are still using `k8s.gcr.io/sig-storage/csi-node-driver-registrar` image , so the validation can not be deleted

ymmt2005 (changes requested):
Please reconfirm the images.

neco-admission/base/deployment.yaml
llamerada-jp: We are confirmed that and we agree with you.
ref. https://github.com/cybozu-go/neco/issues/940

binoue:
No images containing `datadog/agent`, `docker.io/datadog/agent`, and `uay.io/integreatly/grafana_plugins_init` exists in our staging and production environments

ymmt2005 (approved):
LGTM

PR #1409, opened 2021-04-16
topolvm: change socket path to make it be the same as upstream

toshipp:


llamerada-jp (approved):
LGTM

PR #1410, opened 2021-04-16
Changed the alert time threshold depending on whether rolling reboot is in progress or not.

umezawatakeshi:


morimoto-cybozu (approved):
LGTM

PR #1418, opened 2021-04-20
logging: add test to get logs from boot

toshipp:


toshipp:
Waiting until the deployment of promtail on the boot server.

llamerada-jp (approved):
LGTM

PR #1420, opened 2021-04-21
prometheus-adapter: narrow down metrics retrieval

umezawatakeshi:
In order to avoid `__name__!~"^container_.*"` queries, which try to return too much metrics.

The upstream helm chart has no ability to enable `__name__=~"^container_.*"` queries and `__name__!~"^container_.*"` queries separately. We disable all of them and write `__name__=~"^container_.*"` queries as custom queries manually.


tapih (approved):
LGTM

ymmt2005 (approved):
LGTM

PR #1428, opened 2021-04-23
loki: enlarge disk size to avoid disk full for querier

toshipp:
querier caches index whose size is 500Mi per day.
If searches are executed for 20 days, querier requires 10Gi, this makes it disk full.
To prevent such a situation, I enlarge disk space to 100Gi.

### DIFF
--- a/logging/base/loki/upstream/apps-v1.StatefulSet-querier.yaml
+++ b/logging/base/loki/upstream/apps-v1.StatefulSet-querier.yaml
@@ -76,5 +76,5 @@ spec:
       - ReadWriteOnce
       resources:
         requests:
-          storage: 10Gi
+          storage: 100Gi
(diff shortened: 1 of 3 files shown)


llamerada-jp (dismissed):
LGTM

logging/base/loki/upstream/apps-v1.StatefulSet-querier.yaml
```diff
@@ -76,5 +76,5 @@ spec:
       - ReadWriteOnce
       resources:
         requests:
-          storage: 10Gi
```
ymmt2005, line 79: Unfortunately, volumeClaimTemplates in StatefulSet cannot be updated.
You need to update PVC manually, or remove StatefulSet to recreate it.

llamerada-jp (approved):
LGTM

PR #1430, opened 2021-04-23
Disable PodSecurityPolicy

zoetrope:
Remove Unnecessary resources as PSP was disabled.


binoue (dismissed):
Some generated manifests by kustomize still contains PSPs.
(for example, topolvm and victoria metrics operator).

These PSPs are generated by the upstream manifests and expected to be deleted by the upstreams after all.
So I think we do not need to remove them by modifying our `kustomization.yaml` with a lot of effort.

@zoetrope @masa213f 

What do you think?

zoetrope:
@binoue 
Helmed TopoLVM would be able to remove the PSP.

victoriametrics-operator dynamically creates PSPs.
It will need to be fundamentally addressed.

binoue (approved):
LGTM

PR #1474, opened 2021-05-13
use commit hash to stabilize test revision

toshipp:
If the local branch name is "main", argocd sync revision in a test environment refers to remote "main" branch.
If the local branch revision differs from the remote one, the user intended test revision also differs.
To prevent such a situation, use commit hash instead of the branch name.


ymmt2005 (approved):
LGTM

PR #1492, opened 2021-05-19
Updates for k8s 1.20

ysksuzuki:
- Upgrade tools' versions in test/Makefile
- Use `node-role.kubernetes.io/control-plane` instead of `node-role.kubernetes.io/master` that is referred to test and squid.yaml
- Set Service.spec.allocateLoadBalancerNodePorts=false to Load Balancer Services that don't need node port.


morimoto-cybozu (approved):
LGTM

PR #1498, opened 2021-05-20
monitoring: lengthen the retention period for smallset.

toshipp:
The previous value 15d was a bit shorter for post-investigation.
I assume 2 months is sufficient and expect it consumes about 20GB
of disk space but it could not exhaust all space.


ymmt2005 (approved):
LGTM

PR #1499, opened 2021-05-20
Add reboot node time panel

binoue:
Currently, there is no panel to check the number of reboot for each nodes.
As viewing the number is important because kernel panic leads to reboot its node, this PR add the panel to show the reboot timing.

The timing is calculated by the equation
```
deriv(time() - node_boot_time_seconds{}) < bool 0.1
```

### DIFF
--- a/monitoring/base/grafana-operator/dashboards/all-nodes.yaml
+++ b/monitoring/base/grafana-operator/dashboards/all-nodes.yaml
@@ -719,7 +719,7 @@ spec:
                 "steppedLine": false,
                 "targets": [
                     {
-                        "expr": "(deriv(time() - node_boot_time_seconds{}) < bool 0.1)",
+                        "expr": "deriv(time() - node_boot_time_seconds{}) < bool 0.1",
                         "interval": "",
                         "legendFormat": "",
                         "refId": "A"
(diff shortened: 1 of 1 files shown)


umezawatakeshi (approved):

morimoto-cybozu:
@binoue 
Why does the diff contain so many lines unrelated to this issue?

binoue:
@morimoto-cybozu 

> Why does the diff contain so many lines unrelated to this issue?

To add the reboot node time panel, I took the live dashboard json.
So, the changes unrelated to this issue are not what I intentionally did.

Probably, grafana updates the parameters when the dashboard is deployed.
( This dashboard is only updated about its queries since the first deploy on 2020/04/28, so it would be outdated
https://github.com/cybozu-go/neco-apps/commits/main/monitoring/base/grafana-operator/dashboards/all-nodes.yaml )

I thought having the outdated dashboard was not good, but it might be better to add the new panel and keep the rest as it was?

morimoto-cybozu:
@binoue 
The better way would be to split commits.

1. Export the dashboard without adding a new panel.  If this does not produce diff, make a slight change and export.
2. Add a new panel and export.

This is hindsight.  It's OK for this PR to contain unrelated lines, I think.

PR #1515, opened 2021-05-25
Mount .kube directory by tmpfs

binoue:
`kubectl` is sometimes very slow due to its caching.
To speed it up, this PR mounts `.kube` directory by tmpfs.


ymmt2005 (commented):
@binoue 
Why do you choose neco-apps instead of neco repository?
I think neco is more appropriate.

binoue:
Mounting .kube directory by tmpfs should be done in neco, so close this PR.

PR #1517, opened 2021-05-25
Restart teleport-proxy periodically

zoetrope:
There is a problem that teleport does not reload certificates.
https://github.com/gravitational/teleport/issues/3815

We will restart teleport periodically to avoid this problem.

### DIFF
--- a/team-management/template/teleport/base/kustomization.libsonnet
+++ b/team-management/template/teleport/base/kustomization.libsonnet
@@ -9,6 +9,7 @@ function(teams) [{
     'statefulset.yaml',
     'apps',
     'nodes',
+    'restart',
   ],
   configMapGenerator: [
     {
--- a/teleport/base/restart/cronjob.yaml
+++ b/teleport/base/restart/cronjob.yaml
@@ -1,0 +1,23 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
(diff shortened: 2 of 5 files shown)


teleport/base/restart/cronjob.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: restart-teleport-proxy
+spec:
+  concurrencyPolicy: Forbid
+  schedule: "* * * 1 *"
```
masa213f, line 7: This schedule means “At every minute in January.”.
https://crontab.guru/#*_*_*_1_*

https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax
It is good to use `@monthly`.

masa213f (changes requested):

team-management/template/teleport/base/kustomization.libsonnet
```diff
@@ -9,6 +9,7 @@ function(teams) [{
     'statefulset.yaml',
     'apps',
     'nodes',
+    'restart',
```
masa213f, line 12: Why is this change needed?

masa213f (approved):
LGTM

PR #1528, opened 2021-05-27
Modify the AddressPool setting of domestic-egress

ysksuzuki:
Modify the AddressPool setting of domestic-egress for dctest to be the same as stage because it does not match the GlobalNetWorkPolicy settings.

This PR disables upgrade tests temporarily because AddressPool change is rejected in upgrade tests.


yokaze (approved):
LGTM

PR #1541, opened 2021-06-01
update ceph witch comfigmap to avoid bluefs corruption

llamerada-jp:
We found an issue that the prepare job is failed and the osd process does not come up.
The issue is related to bluefs io mode, so to avoid the issue, we set bluefs_buffered_io to alse.


toshipp (approved):
LGTM

PR #1545, opened 2021-06-01
Rook: use `.spec.resources.osd` to set resource limits for OSD

llamerada-jp:
The field `.spec.resources.osd` was not worked but it has fixed in Rook
1.6 or lator. Set the resource limits using the `.spec.resources` field in line with other resources.


toshipp (approved):
LGTM

PR #1554, opened 2021-06-03
fix reboot count query

binoue:
Sometimes, `node_boot_time_seconds` fluctuates in a small range especially when its node is up.
Such fluctuations are nothing related to rebooting, so fix the query for detecting rebooting nodes to ignore them.


morimoto-cybozu (approved):
LGTM

PR #1561, opened 2021-06-04
teleport: add node pod for neco team 

yokaze:
This PR allows neco team to use teleport node pod.
It deploys:
- a StatefulSet for the node pod,
- a ServiceAccount supplied to the node pod,
- a ClusterRoleBinding that binds `cluster-admin` privilege to the ServiceAccount.

As a result, users accessing the teleport node can operate on the k8s cluster as `cluster-admin`.
<br />

Since the neco node pod has access to powerful privilege, it is important to restrict who can access it.
Neco clusters manage this access control using teleport node labels.
The authorization flow is:
1. users are mapped to (one or multiple) teleport roles regarding the secret settings,
2. a role (e.g. [admin](https://github.com/cybozu-go/neco-apps/blob/main/teleport/base/conf/admin-role.yaml#L14)) allows users to access teleport nodes with matching `team` labels,
3. each node are configured with `team` labels and are accessible to users (See [teleport-node.service](https://github.com/cybozu-go/neco/blob/main/debian/lib/systemd/system/teleport-node.service#L15) and [ept node pod](https://github.com/cybozu-go/neco-apps/blob/main/teleport/base/nodes/ept.yaml#L33)).

Neco team members are already configured for accessing teleport nodes with `team=neco` label, because it is used for accessing boot servers.
Therefore, the access control configuration is not included in this PR.


zoetrope:
Please fix conflicting.

masa213f (approved):

zoetrope (approved):

PR #1579, opened 2021-06-10
topolvm: add new storageclasses

toshipp:
This introduces two new storageclasses to provide ext4 and xfs using topolvm.
Previously, a SC with xfs using topolvm is provided as topolvm-provisioner,
but its name is not user friendly, so we add a more sophisticated one and
retain topolvm-provisioner for compatibility.

### DIFF
--- a/topolvm/base/kustomization.yaml
+++ b/topolvm/base/kustomization.yaml
@@ -12,13 +12,14 @@ resources:
 - upstream/overlays/daemonset-scheduler/scheduler.yaml
 - pdb.yaml
 - priorityclass.yaml
+- storageclass.yaml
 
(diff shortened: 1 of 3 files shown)


satoru-takeuchi:
@toshipp Why do you put storage classes into two separate files?

toshipp:
@satoru-takeuchi
I prefer grouping storageclasses for better understanding structure.
If separate them, many manifests may exist in the directory in the future.
Possibly creating a directory for sc is better but now it is a bit over-engineering, IMO.

satoru-takeuchi:
@toshipp I understood your thought, thank you.

satoru-takeuchi (approved):

PR #1584, opened 2021-06-11
save promtail's positions file to /var/log

umezawatakeshi:


umezawatakeshi:
abandon

PR #1587, opened 2021-06-14
Add base hnc configuartion

akazs:
This PR adds HNC only for staging and test environments.

HNC has a functionality to convert an existing name to a child namespace, but all tenant teams can not use it because of the following reasons
1. If the target namespace is **not** under the team's control, it can not make the namespace as a child namespace beacause of not having the create permission of `hierarchyconfigurations` in the namespace.
2. If the target namespace is under the team's control, then Role and RoleBinding must be propagated from the parent namespace but it fails as the same named RoleBinding already exists in our environment.

Therefore, this PR does not have the test case to check if tenant teams can create child namespaces.


argocd-config/base/hnc.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: argoproj.io/v1alpha1
```
zoetrope, line 1: Is this file being used?

hnc/base/clusterroles/cluster_role_aggregate_hnc_admin_edit.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: hnc-admin-role
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true" 
+    rbac.authorization.k8s.io/aggregate-to-edit: "true" 
+rules:
+- apiGroups:
+  - hnc.x-k8s.io
+  resources:
+  - hierarchyconfigurations
```
zoetrope, line 12: I don't think tenants should have access to hierarchyconfigurations.

hnc/base/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - --excluded-namespace=app-gorush
+        - --excluded-namespace=app-kodama
+        - --excluded-namespace=app-maneki-static-cybozu-com
+        - --excluded-namespace=app-misc
+        - --excluded-namespace=app-monitoring
+        - --excluded-namespace=app-monitoring-elasticstack
+        - --excluded-namespace=app-oauth-redirector
+        - --excluded-namespace=app-octodns
+        - --excluded-namespace=maneki
+        - --excluded-namespace=maneki-dev
```
zoetrope, line 69: Why is maneki-dev excluded?

hnc/base/hnc-system.yaml
```diff
@@ -0,0 +1,5 @@
+$patch: delete
```
zoetrope, line 1: Why do you want to remove hnc-system namespace?

binoue: Because `hnc-system` namespace is created by namespace Application
https://github.com/cybozu-go/neco-apps/pull/1587/files#diff-c9c30507c8cbf8c47f767b445bb9e5d43552cd121be6c4250ea726bd6ee398d8

test/hnc_test.go
```diff
@@ -0,0 +1,135 @@
...
+metadata:
+  name: %s
+  namespace: %s
+`
+
+	By("creating subnamespace with subnamespaceanchor")
+	stdout, stderr, err := ExecAtWithInput(boot0,
+		[]byte(fmt.Sprintf(subnamespaceAnchorYAMLtemplate,
+			HNCTestSubnamespaces[0], HNCTestNamespace)),
+		"kubectl", "apply", "--as-group=tenant", "--as=tenant", "-f", "-")
```
zoetrope, line 58: Do you need `--as-group=system:authenticated`?

test/hnc_test.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil
+	}).Should(Succeed())
+}
+
+func deleteSubnamespace() {
+	By("deleting subnamespaceanchor")
+	stdout, stderr, err := ExecAt(boot0,
+		"kubectl", "delete", "--as-group=tenant", "--as=tenant",
+		"subnamespaceanchors.hnc.x-k8s.io", HNCTestSubnamespaces[1], "-n", HNCTestNamespace)
+	Expect(err).ShouldNot(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
```
zoetrope, line 126: Don't you check for the existence of subnamespace?

hnc/base/hnc-system.yaml
zoetrope: Why?

zoetrope: We used to use namespaces application to prevent the deletion of namespaces.
Now we don't have that problem, so we can create namespaces for each application.

test/validation_test.go
```diff
@@ -712,6 +714,78 @@ func testVMCustomResources(t *testing.T) {
...
+		}
+
+		expectedExcludedNamespaces = append(expectedExcludedNamespaces, meta.Name)
+	})
+	for _, v := range expectedExcludedNamespaces {
+		if _, ok := actualExcludedNamespaces[v]; !ok {
+			t.Errorf("Namespae %v is not actually excluded", v)
+		}
+	}
+}
```
masa213f, line 787: Why don't you check by `cmp.Equal`?
lack of `hnc.x-k8s.io/excluded-namespace` label cannot be detected, I think.

hnc/base/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - --excluded-namespace=logging
+        - --excluded-namespace=metallb-system
+        - --excluded-namespace=moco-system
+        - --excluded-namespace=monitoring
+        - --excluded-namespace=prometheus-adapter
+        - --excluded-namespace=psa-system
+        - --excluded-namespace=pvc-autoresizer
+        - --excluded-namespace=registry
+        - --excluded-namespace=teleport
+        - --excluded-namespace=topolvm-system
```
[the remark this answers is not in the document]
zoetrope: I'm worried about forgetting to maintain the list.
Can you consider automatic generation by jsonnet?

hnc/base/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - --excluded-namespace=app-garoon-static
+        - --excluded-namespace=app-gorush
+        - --excluded-namespace=app-kodama
+        - --excluded-namespace=app-maneki-static-cybozu-com
+        - --excluded-namespace=app-misc
+        - --excluded-namespace=app-monitoring
+        - --excluded-namespace=app-monitoring-elasticstack
+        - --excluded-namespace=app-oauth-redirector
+        - --excluded-namespace=app-octodns
+        - --excluded-namespace=maneki
```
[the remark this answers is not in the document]
zoetrope: We also have to check the namespace of neco-tenant-apps.
Umm...

namespaces/base/argocd.yaml
```diff
@@ -4,3 +4,4 @@ metadata:
   name: argocd
   labels:
     pod-security.cybozu.com/policy: privileged
+    hnc.x-k8s.io/excluded-namespace: "true"
```
zoetrope, line 7: Is it possible to use commonLabels like team label?

test/validation_test.go
binoue: `hnc.x-k8s.io/excluded-namespace` does not affect excluding namespaces.
The situation we should avoid is that we expect to exclude some namespaces by the label on them, but actually we don't.

But yes, using `cmp.Equal` might be better because we can notice lacking of labels too.

hnc/base/clusterroles/cluster_role_aggregate_hnc_admin_edit.yaml
binoue: I did not find any problem about giving the edit permission for hierarchyconfigurations to tenant teams because the edit operation would be blocked by the HNC webhook like below.
```
error: hierarchyconfigurations.hnc.x-k8s.io "hierarchy" could not be patched: admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Illegal parent: Cannot set the parent of "test-2" to "test1" because it's a subnamespace of "test2"
You can run `kubectl replace -f /tmp/kubectl-edit-rzpyz.yaml` to try this update again.
```

But I confirmed that the edit permission for hierarchyconfigurations is not necessary to use HNC, so I fix the permission

test/hnc_test.go
binoue: fixed

binoue:
We will use [Innu](https://github.com/cybozu-go/innu) in future, so close this PR.

PR #1594, opened 2021-06-17
parse API server port too from ~/.kube/config

umezawatakeshi:
It would be nice to parse not only API server hostname but also its port number from `~/.kube/config`. The port number may be changed by `ckecli kubernetes issue`.


morimoto-cybozu (approved):
LGTM

Issue #1477, opened 2021-05-13
[GCP] Change `sandbox` namespace's Pod privilege to `baseline`


ymmt2005:
## What

On a GCP environment, `sandbox` namespace allows to run privileged Pods.
This differs from our real data center environments and makes things difficult to debug.

Use another namespace for tests, and have `sandbox` restricted.

## How

Describe how to address the issue.

## Checklist

- [x] Finish implentation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #1597, opened 2021-06-17
use another namespace for dctest and drop privilege from sandbox namespace

umezawatakeshi:
part of https://github.com/cybozu-go/neco-apps/issues/1477


yokaze (approved):
LGTM

PR #1614, opened 2021-06-23
delete registry namespace

yokaze:


zoetrope (approved):

zoetrope:
Is the actual deletion performed as a manual operation?

yokaze:
@zoetrope I think we need to annotate the namespace to proceed the deletion.

PR #1627, opened 2021-06-25
monitoring: improve ceph dashboard

toshipp:
- make dashboards readonly
- make namespace variable be refreshed when the dashboard is loaded
- fix OSD status when showing long term


llamerada-jp (approved):
LGTM

PR #1630, opened 2021-06-28
change argocd-config's sync-wave to 9999

umezawatakeshi:
Change 'argocd-config' Application's sync-wave in order to sync this Application last.

Before this PR, argocd-config's sync-wave is not set (= sync-wave "0"). This may cause syncing other child Applications themselves (such as adding/deleting Applications) stuck.

see also: https://github.com/cybozu-go/neco/issues/1661


zoetrope (approved):
LGTM

PR #1641, opened 2021-06-30
rook: double mgr pods

toshipp:
- double mgr pods to high availability
- add anti-affinity to spread across racks


llamerada-jp (approved):
LGTM

PR #1652, opened 2021-07-02
Disable hash suffix for VMAlertmanager ConfigMaps to stabilize update

morimoto-cybozu:
This commit disables the hash suffix functionality for the ConfigMaps
used in VMAlertmanager CRs.
The update of ConfigMap contents is detected by the configmap-reload
sidecar containers instead.
With this commit we can avoild the early deletion of old ConfigMaps
by Argo CD, the instability of the VMAlertmanager StatefulSets, and
the hang-up of VictoriaMetrics operator's reconciliation.

### DIFF
--- a/monitoring/base/victoriametrics/vmalertmanager-largeset.yaml
+++ b/monitoring/base/victoriametrics/vmalertmanager-largeset.yaml
@@ -8,20 +8,41 @@ spec:
   # VMAlertmanager is able to detect configSecret content change and to reload alertmanager.
   # We don't need to use secretGenerator for alertmanager.yaml unlike monitoring/base/alertmanager/deployment.yaml
   configSecret: vmalertmanager-config
+  # However, it is not able to detect configMap content change. So use another configmap-reload container.
+  # Note that we don't use VMAlertmanager.spec.configMaps because they use name of ConfigMap as mount point.
+  # We need a common mount point for largeset and smallset.
   volumeMounts:
     - name: alertmanager-config-volume
       mountPath: /etc/vm/configs/vmalertmanager
     - name: vmalertmanager-settype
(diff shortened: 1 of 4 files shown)


monitoring/base/victoriametrics/vmalertmanager-largeset.yaml
```diff
@@ -8,20 +8,41 @@ spec:
...
+        - -web.listen-address=:9534
+      volumeMounts:
+        - name: alertmanager-config-volume
+          mountPath: /etc/vm/configs/vmalertmanager
+          readOnly: true
+        - name: vmalertmanager-settype
+          mountPath: /etc/vm/configs/vmalertmanager-settype
+          readOnly: true
+      resources:
+        limits:
```
masa213f, line 43: Sorry to ask a minor thing. Why do you set `resources.limits` only?
If you intended to set the same value of the config-reloader sidecar container which created VMOperator, it is good to set both `resources.limits` and `resources.requests`, I think.

morimoto-cybozu: @masa213f 
As you wrote, I intended to set the same values here as the operator sets.  The operator sets only the limits in StatefulSet and I'm doing the same.
K8s works fine even without the requests.  I suppose that the StatefulSet controller fills the requests for the Pods, as seen in the vmalertmanager Pods.

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits
> If a Container specifies its own memory limit, but does not specify a memory request, Kubernetes automatically assigns a memory request that matches the limit.

masa213f: Thank you. I did not know the behavior.

masa213f (approved):

PR #1658, opened 2021-07-07
Rook: enable PDB for ceph daemons

toshipp:
- Rook manages PDB for OSD, MON, RGW.
  - If the number of RGW is two, Rook does not create PDB (BUG?), so we create 3 instances for a workaround.
- Current Rook does not manage PDB for MGR, so we create it manually.

issue: cybozu/csa#3


llamerada-jp (approved):
LGTM

llamerada-jp (approved):
LGTM

PR #1669, opened 2021-07-12
Rook/Ceph: disable pdb for osd to avoid blocking node reboot

llamerada-jp:


toshipp (approved):
LGTM

PR #1704, opened 2021-08-04
add Accurate

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1704

- add [Accurate](https://github.com/cybozu-go/accurate)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,16 @@ JSONNET_LIBS_K8S_ALPHA_VERSION = 1.20
 all:
 	@echo Read docs/maintenance.md for the usage
 
+.PHONY: update-accurate
+update-accurate:
+	$(call get-latest-gh,cybozu-go/accurate)
+	rm -rf /tmp/accurate
+	cd /tmp; git clone --depth 1 -b $(latest_gh) https://github.com/cybozu-go/accurate
+	rm -rf accurate/base/upstream/*
+	cp -a /tmp/accurate/config/* accurate/base/upstream
+	rm -rf /tmp/accurate
+	sed -i -E 's/newTag:.*$$/newTag: $(subst v,,$(latest_gh))/' accurate/base/kustomization.yaml
+
 .PHONY: update-argocd
 update-argocd:
 	$(call get-latest-tag,argocd)
--- a/accurate/base/namespace.yaml
+++ b/accurate/base/namespace.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: system
+  name: accurate
   labels:
     team: neco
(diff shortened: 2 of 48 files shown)


Makefile
```diff
@@ -8,6 +8,16 @@ JSONNET_LIBS_K8S_ALPHA_VERSION = 1.20
...
 
+.PHONY: update-accurate
+update-accurate:
+	$(call get-latest-gh,cybozu-go/accurate)
+	rm -rf /tmp/accurate
+	cd /tmp; git clone --depth 1 -b $(latest_gh) https://github.com/cybozu-go/accurate
+	rm -rf accurate/base/upstream/*
+	cp -a /tmp/accurate/config/* accurate/base/upstream
+	rm -rf /tmp/accurate
+	sed -i -E 's/newTag:.*$$/newTag: $(subst v,,$(latest_gh))/' accurate/base/kustomization.yaml
```
ymmt2005, line 19: There's no need to replace the image tag because Accurate is owned by us.

ymmt2005: Ah, the file is not from the upstream.

accurate/base/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: system
```
ymmt2005, line 4:
```suggestion
  name: accurate
```

Makefile
umezawatakeshi: The upstream sets image tag by [kustomization.yaml at repo root](https://github.com/cybozu-go/accurate/blob/e70599c62a283a4c05dfbe83cf2658022a40e008/kustomization.yaml). This kustomization.yaml also generates configMap but its content should be modified in neco-apps. So we cannot use the kustomization.yaml and write `newTag` by ourselves.

accurate/base/namespace.yaml
umezawatakeshi: If so, the Namespace won't be patched.

... I checked `kustomize build` result of original kustomization.yaml, the name of the Namespace becomes `accurate-system`, not `accurate`. ...!?

ymmt2005: This is my result.

```console
$ kustomize build . | head
apiVersion: v1
kind: Namespace
metadata:
  labels:
    app.kubernetes.io/name: accurate
  name: accurate
```

ymmt2005: I guess your `kustomize` is an old version.

```
$ kustomize version
{Version:kustomize/v4.1.3 GitCommit:0f614e92f72f1b938a9171b964d90b197ca8fb68 BuildDate:2021-05-20T20:52:40Z GoOs:linux GoArch:amd64}
```

umezawatakeshi: Oops, got it.

ymmt2005 (approved):
LGTM

PR #1710, opened 2021-08-11
add meows

kmdkuk:


meows/overlays/gcp/namespace.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: meows
+  labels:
+    team: neco
+    pod-security.cybozu.com/policy: privileged
```
zoetrope, line 7: Why do you need `privileged`?
I don't think controller and slack-agent need `privileged`.

monitoring/base/victoriametrics/rules/meows-alertrule.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: operator.victoriametrics.com/v1beta1
+kind: VMRule
+metadata:
+  name: meows
+  namespace: monitoring
+spec:
+  groups:
+    - name: meows
+      rules:
+        - alert: MeowsDown
```
zoetrope, line 10: Isn't this the same as KubePodNotReady?

zoetrope (approved):

PR #1729, opened 2021-08-23
Delete pvc-autoresizer overlay for stage-0.

cupnes:
issue: cybozu/csa#55


toshipp (approved):

llamerada-jp (approved):
LGTM

llamerada-jp (commented):
Would you consider adding some tests to detect a similar settings?
For example, in helm applications, if targetRevision is stage/release, it will cause an test failure.

toshipp:
@llamerada-jp 
I think such tests are good, but if we use helm chart in neco-apps in the future, the tests will fail.
So, it would be hard to maintain generalized validation tests, IMO.
Anyway, this fix is required to recover our stage environment, let's discuss it in another issue.

PR #1743, opened 2021-08-28
Disable webhooks for psa-system namespace not to prevent bootstrap

morimoto-cybozu:


ymmt2005:
I'd like to deploy PSA in `kube-system` as it should be considered a system service.

meows/overlays/gcp/namespace.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
   labels:
     team: neco
     meows.cybozu.com/pod-mutate: ignore
+    admission.cybozu.com/pod: ignore
+    topolvm.cybozu.com/webhook: ignore
+    pod-security.cybozu.com/policy: privileged
```
ymmt2005, line 10: I don't think these are necessary.

namespaces/base/topolvm.yaml
```diff
@@ -6,3 +6,5 @@ metadata:
     app.kubernetes.io/name: topolvm-system
     topolvm.cybozu.com/webhook: ignore
     pod-security.cybozu.com/policy: privileged
+    admission.cybozu.com/pod: ignore
+    meows.cybozu.com/pod-mutate: ignore
```
ymmt2005, line 10: ditto

meows/overlays/gcp/namespace.yaml
ymmt2005: Also, I'm curious about the reason why we are using `namespace` app for meows Namespace.

morimoto-cybozu:
I'm closing this PR because I have changed the strategy radically in #1753.

PR #1753, opened 2021-09-01
Refine meows and pod-security-admission webhooks

morimoto-cybozu:
This PR refines the following webhooks.
* meows
  * Limit the target namespaces of the webhook
* pod-security-admission
  * Import the latest manifests from the upstream
    * The PSA objects are deployed into kube-system namespace instead of psa-system namespace
    * This prevents the PSA Pods from being blocked by webhooks


ymmt2005 (approved):
LGTM

PR #1763, opened 2021-09-10
Set accurate template for tenant root namespaces

yokaze:
This PR adds `init-template` Application that deploys `init-template` namespace.
It is annotated as a template namespace of Accurate, that stores common resources.

### DIFF
--- a/argocd-config/base/init-template.yaml
+++ b/argocd-config/base/init-template.yaml
@@ -1,0 +1,23 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: init-template
+  namespace: argocd
+  annotations:
+    argocd.argoproj.io/manifest-generate-paths: ..
+    argocd.argoproj.io/sync-wave: "7"
+  finalizers:
+  - resources-finalizer.argocd.argoproj.io
+spec:
+  project: default
+  source:
+    repoURL: https://github.com/cybozu-go/neco-apps.git
+    targetRevision: release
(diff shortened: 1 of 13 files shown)


accurate/base/tenant-template.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
```
ymmt2005, line 1: accurate アプリにふくまれるべきではないと思います。namespaces 辺りでは。

accurate/base/kustomization.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - upstream/default
+- tenant-template.yaml
```
ymmt2005, line 5: ditto

accurate/base/tenant-template.yaml
yokaze: @ymmt2005 namespaces の下だと `team: neco` が付いてしまうみたいですね。。それを避けようとすると新しいアプリを作る必要があるかもです。。

ymmt2005 (approved):
Seems OK, but I have a question.

Will namespaces such as `app-kodama` become a sub-namespace of Accurate?

yokaze:
@ymmt2005 I think such namespaces may be the root namespaces derived from `init-template`, but I will investigate the appropriate strategy for it.

PR #1782, opened 2021-09-21
fix VMPodScrapes for moco

umezawatakeshi:
It seems that the original manifest selects all pods.


morimoto-cybozu (approved):
LGTM

PR #1787, opened 2021-09-27
fix KubeJobFailed

yamatcha:
This PR add KubeOwnerJobFailed alert rule to check the latest job if the Job has owner(controller, CronJob ,etc...).
And I add KubeJobFailed alert rule to check the job without owner.

Signed-off-by: yamatcha <soju-yamashita@cybozu.co.jp>

### DIFF
--- a/monitoring/base/victoriametrics/rules/kubernetes-alertrule.yaml
+++ b/monitoring/base/victoriametrics/rules/kubernetes-alertrule.yaml
@@ -446,21 +446,21 @@ spec:
           for: 1h
           labels:
             severity: warning
-        - alert: KubeOwnerJobFailed
+        - alert: KubeCronJobFailed
           annotations:
-            summary: OwnerJob {{ $labels.namespace }}/{{ $labels.owner_name }} failed to complete.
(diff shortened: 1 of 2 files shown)


monitoring/base/victoriametrics/rules/kubernetes-alertrule.yaml
```diff
@@ -446,12 +446,34 @@ spec:
           for: 1h
           labels:
             severity: warning
+        - alert: KubeOwnerJobFailed
```
zoetrope, line 449: I don't think `OwnerJob` is appropriate name.
How about `KubeCronJobFailed`?

yamatcha: I think `KubeOwnerJobFailed` is appreciate rather than `KubeCronJobFailed`.
Because the target job is not only CronJob, but also the Job created by controller.
But I don't think `KubeOwnerJobFailed` is not best.
I have the alternative `KubeJobWithOwnerFailed`, but it is also unclear and long.

zoetrope (approved):
LGTM

PR #1798, opened 2021-09-30
Remove garoon namespaces

yokaze:


team-management/base/garoon/elastic-serviceaccount.yaml
```diff
@@ -7,18 +7,6 @@ metadata:
...
-kind: ServiceAccount
-metadata:
-  name: elastic
-  namespace: app-garoon-static
----
-apiVersion: v1
-kind: ServiceAccount
 metadata:
   name: elastic
   namespace: dev-garoon
```
zoetrope, line 12: We can remove elastic service account completely.
Because app-garoon and dev-garoon does not use the elastic service account.

zoetrope (approved):
LGTM

PR #1807, opened 2021-10-04
Strict image validation

masa213f:


kmdkuk (commented):
It would be difficult to make sure that you can't pull all the images that are not allowed, but how about including one test to make sure (e.g. https://hub.docker.com/_/ubuntu).

zoetrope (dismissed):
LGTM

kmdkuk (approved):
LGTM

PR #1813, opened 2021-10-04
use .spec.ingressClassName in HTTPProxy test

umezawatakeshi:
forgot to update test code.

### DIFF
--- a/test/contour_test.go
+++ b/test/contour_test.go
@@ -143,6 +143,7 @@ func testContour() {
 		fqdnHTTP := testID + "-http.test-ingress.gcp0.dev-ne.co"
 		fqdnHTTPS := testID + "-https.test-ingress.gcp0.dev-ne.co"
 		fqdnBastion := testID + "-bastion.test-ingress.gcp0.dev-ne.co"
+		fqdnBastionAnnotated := testID + "-bastion-annotated.test-ingress.gcp0.dev-ne.co"
 
 		By("getting contour service")
 		var targetIP string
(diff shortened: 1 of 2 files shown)


morimoto-cybozu:
@umezawatakeshi 
This data file is just for creating HTTPProxy, not for testing the admission webhook around HTTPProxy.  And there are many other test-data files which uses `kubernetes.io/ingress.class` annotation.
Although updating data files to a newer format has its own worth, updating `admission_test.go` would be appropriate to update the tests about the admission webhook.

umezawatakeshi:
Tests around admission webhook is already updated in https://github.com/cybozu-go/neco-apps/pull/1780 (https://github.com/cybozu-go/neco-apps/pull/1780/commits/d3c60aabff9da0a2ff3051381fadee88f45ccb27). So "forgot to update test **code**." may not a correct wording...

Anyway, I will update test-data files to use `.spec.ingressClassName`. Is it OK?

morimoto-cybozu:
@umezawatakeshi 
I'll make an explicit reply.

> Tests around admission webhook is already updated in #1780 (d3c60aa).

Ah, I overlooked it.

> Anyway, I will update test-data files to use `.spec.ingressClassName`. Is it OK?

Sure, please.

umezawatakeshi:
@morimoto-cybozu 

As I was modifying the test-data, I noticed that the `HTTPProxy`s in test-data must declare ingress classes in the same way as the one in the main manifests to test correctly. So now I don't think I should change them to use `.spec.ingressClassName` fields.

Currently, neco-admission doesn't support the migration from `kubernetes.io/ingress.class` annotation to `.spec.ingressClassName` field, so we can't change the main manifests either.

morimoto-cybozu:
@umezawatakeshi 
Well, will you close this PR?  Or freeze this until neco-admission supports migration?

umezawatakeshi:
@morimoto-cybozu 
I still wonder whether we should update manifests used by contour_test.go. What do you think?

morimoto-cybozu:
@umezawatakeshi 
If the set of the manifests used by contour_test.go is a subset of the production manifests set, we cannot update them until neco-admission supports migration, as you wrote.

It the manifests used by contour_test.go are independent of the production manifests, we can either update them immediately or defer update.
Anyway we should update them at some future time because the annotation style will be deprecated and may be removed eventually.

morimoto-cybozu (approved):
LGTM

PR #1841, opened 2021-10-08
extend disk size used `/var/scratch` in dctest-runner

kmdkuk:
If you let dctest-runner wait for a long time, the size of the volume prepared in `/var/scratch` will be exhausted and the test execution environment will die.
I know that the 300Gi size that I have available will be consumed in about 20 hours at this point.
It consumes about 100Gi when it is just started, and the trend is roughly linear, consuming 100Gi per 10 hours.
If this trend remains unchanged for more than 20 hours, this change will increase the deadline of the test execution environment by 10 hours.
In the future, meows will be added a feature to recreate the test execution environment.
We plan to use the feature to recreate the test execution environment in 24 hours, so 30 hours of its availability will be sufficient.
While the feature to recreate cannot be enabled, you will need to manually delete the dead test environment.

### DIFF
--- a/meows/overlays/stage0/runnerpool.yaml
+++ b/meows/overlays/stage0/runnerpool.yaml
@@ -60,7 +60,7 @@ spec:
                 accessModes: [ "ReadWriteOnce" ]
                 resources:
                   requests:
-                    storage: 300Gi
+                    storage: 400Gi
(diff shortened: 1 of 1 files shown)


zoetrope (approved):
LGTM

PR #1850, opened 2021-10-13
Revert "Remove cydec and maneki namespaces"

yokaze:
Reverts cybozu-go/neco-apps#1839


test/setup_test.go
```diff
@@ -353,31 +353,16 @@ func testSetup() {
 }
 
 func applyAndWaitForApplications(commitID string) {
-	// TODO: remove this block after #1839 is released
-	By("grafting cydec namespaces")
+	// TODO: remove this block after #1798 is released
```
zoetrope, line 356: I think #1798 has already been merged.
Do we need this block?

yokaze: As mentioned, it is not needed.
I would like to revert this as soon as possible, so can I remove it with another PR?

zoetrope: I see. Let's do it.

zoetrope (approved):

PR #1876, opened 2021-10-19
Replace Loki backend from consul to memberlist

zoetrope:


masa213f (commented):
Why don't you add the containerPort setting for the `7946` port to the distributor, ingester and querier?

masa213f (approved):
LGTM

PR #1884, opened 2021-10-20
increase cs nodes

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1793

Signed-off-by: UMEZAWA Takeshi takeshi-umezawa@cybozu.co.jp


morimoto-cybozu (approved):
LGTM

PR #1937, opened 2021-11-02
fix RunnerPool image for test

kmdkuk:
- The image of the RunnerPool resource used for testing was specified with the latest tag.
  - In connection with the release of meows, only the Image version of runner has been increased and is no longer executable.
- Fixed to explicitly specify the version.


morimoto-cybozu (approved):
LGTM

PR #1964, opened 2021-11-08
Move confirmOsdPrepare() to testSetup().

cupnes:
When restarting the node, the pods are not created and confirmOsdPrepare() falls into an error.

This is a problem with the point where confirmOsdPrepare() is called.

Therefore, we changed the location of the call to confirmOsdPrepare().


toshipp (approved):
mobbed

PR #2010, opened 2021-11-16
Rook/Ceph: expand capacity of stage0 ceph-ssd

llamerada-jp:

### DIFF
--- a/rook/overlays/stage0/kustomization.yaml
+++ b/rook/overlays/stage0/kustomization.yaml
@@ -16,4 +16,4 @@ patchesJson6902:
     patch: |-
       - op: replace
         path: /spec/storage/storageClassDeviceSets/0/count
-        value: 23
+        value: 43


satoru-takeuchi (approved):

satoru-takeuchi:
The reason of how Yuji and I choose the new number of OSDs.

Currently, there are 23 OSDs, the total size is 25.3TB, the consumed space is 9.33 TB, and the percentage of free space is about 63.1%. We need to increase the remaining capacity to larger than 70%.

We'd like to set the target value to 80% to prevent frequent operations. Then the additional space becomes 9.33/(1-0.8) ~= 46.64 TB. It means the number OSDs should be 46.64 / 1TiB ~= 43.

At this time, the total space of SSDs in all CSes is `sum(topolvm_volumegroup_size_bytes) = 368,024,724,963,328 Byte` and the total free space is 
 `sum(topolvm_volumegroup_available_bytes) = 207,950,220,099,584 Byte`. After increasing the number of OSDs, the percentage of free space is still 49%. So we regards 43 as a reasonable count.

toshipp (approved):
Understood.

PR #2011, opened 2021-11-16
Rook/Ceph: disable warning by MANY_OBJECTS_PER_PG for ceph-ssd

llamerada-jp:
Suppress the folowing warning.

```
health: HEALTH_WARN
        1 pools have many more objects per pg than average
```

The definition of this parameter is as follows.

https://github.com/ceph/ceph/blob/d31130d4516a36c585b1224c193ec720f31c649c/src/common/options/mgr.yaml.in#L237-L247
```
- name: mon_pg_warn_max_object_skew
  type: float
  level: advanced
  desc: max skew few average in objects per pg
  fmt_desc: Raise ``HEALTH_WARN`` if the average RADOS object count per PG
    of any pool is greater than ``mon_pg_warn_max_object_skew`` times
    the average RADOS object count per PG of all pools. Zero or a non-positive
    number disables this. Note that this option applies to ``ceph-mgr`` daemons.
  default: 10
  services:
  - mgr
```

This warning is harmless in Neco's Ceph clusters because we don't care the difference of the number of objects per pgs between pools. For example, we hit this problem because there is a big pool and there are many almost empty pools.

### DIFF
--- a/rook/base/ceph-hdd/configmap.yaml
+++ b/rook/base/ceph-hdd/configmap.yaml
@@ -7,4 +7,12 @@ metadata:
     argocd.argoproj.io/sync-wave: "-1"
 data:
   # this is patched by overlays, please check overlays if you update it.
-  config: ""
+  config: |
+    [mon]
+    ; Suppress the folowing warning.
+    ;
+    ; ```
+    ; health: HEALTH_WARN
+    ;        1 pools have many more objects per pg than average
+    ; ```
+    mon_pg_warn_max_object_skew = 0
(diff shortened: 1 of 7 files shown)


satoru-takeuchi (dismissed):

toshipp (changes requested):
The warning happens on ceph-object-store.
BTW, I'd like to set this configuration for all clusters because differences may suffer us when troubleshooting.
Let me know what you think.

llamerada-jp:
Indeed. I think we will apply the same configuration if we get a warning on the another cluster.

llamerada-jp:
↑ @satoru-takeuchi 
What do you think?

satoru-takeuchi:
@llamerada-jp I consider that it's better to apply this configuration to all our Ceph clusters as @toshipp said.

toshipp (approved):

PR #2016, opened 2021-11-17
Use argocd app list instead of argocd app get in checkAllAppsSynced

ysksuzuki:
Argocd synchronization process takes so long even after all applications have been synched because each request of `argocd app get` in checkAllAppsSynced takes more than 1 second.

Signed-off-by: Yusuke Suzuki <yusuke-suzuki@cybozu.co.jp>


zoetrope (approved):
LGTM

ymmt2005 (approved):
LGTM

PR #2017, opened 2021-11-17
Update meows v0.4.2

kmdkuk:
- Add a test to check the communication from controller to slack-agent


test/meows_test.go
```diff
@@ -53,24 +55,37 @@ func prepareMeows() {
...
 
 		Eventually(func() error {
 			return checkDeploymentReplicas("slack-agent", "meows", 2)
 		}).Should(Succeed())
+
+		By("getting runner pod list")
+		controllerPodList := new(corev1.PodList)
+		stdout := ExecSafeAt(boot0, "kubectl", "get", "pods", "-n", "meows", "-l=app.kubernetes.io/name=meows,app.kubernetes.io/component=controller", "-o=json")
+		err := json.Unmarshal(stdout, controllerPodList)
+		Expect(err).NotTo(HaveOccurred())
```
masa213f, line 75: This logic is unnecessary. We can use `kubectl exec deploy/moco-controller` :)

test/meows_test.go
```diff
@@ -53,24 +55,37 @@ func prepareMeows() {
...
+
+		By("accessing to slack-agent: should be allow")
+		runnerPoolName := genRunnerPoolName()
+		// This command does not want to check the communication with the slack api, so ignore the error.
+		// Cannot be checked by curl command or other methods.
+		ExecAt(boot0, "kubectl", "exec", "-n", controllerPodList.Items[0].Namespace, controllerPodList.Items[0].Name, "--", "/usr/local/bin/meows", "slackagent", "send", runnerPoolName, "success", "-s", "http://slack-agent.meows.svc", "-c", "test1", "-n", controllerPodList.Items[0].Namespace)
```
masa213f, line 81: Are you confusing the controller's namespace and a runner's namespace?

masa213f (changes requested):

test/meows_test.go
kmdkuk: I understand that.
I'm aware that the Pod name and namespace used here is the string used in the Slack notification.
Therefore, I thought it would be easier to understand the test notifications if I gave the controller namespace when testing.

Since specifying namespace should not be important, I may not need to specify the namespace.

masa213f (approved):
LGTM

PR #2037, opened 2021-11-22
Alert if clock skew is detected

yokaze:
This PR adds two alerts.
- ChronyMetricsMissing: alert if chrony metrics is missing on machines that are visible to sabakan.
  It is expected to find failure when node_exporter is running but chrony_monitor is not rendering metrics correctly.
- NodeClockSkew: alert if node's system time is fastened or delayed by a certain amount.

The severity is set to warning for a while.

### DIFF
--- a/monitoring/base/victoriametrics/rules/node-alertrule.yaml
+++ b/monitoring/base/victoriametrics/rules/node-alertrule.yaml
@@ -62,3 +62,19 @@ spec:
           annotations:
             summary: Available filesystem is low at `{{ $labels.instance }}` on `{{ $labels.mountpoint }}`.
             runbook: Please check filesystem consumption.
+        - alert: ChronyMetricsMissing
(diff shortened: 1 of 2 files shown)


monitoring/base/victoriametrics/rules/node-alertrule.yaml
```diff
@@ -62,3 +62,19 @@ spec:
           annotations:
             summary: Available filesystem is low at `{{ $labels.instance }}` on `{{ $labels.mountpoint }}`.
             runbook: Please check filesystem consumption.
+        - alert: ChronyMetricsMissing
+          expr: sum(sabakan_machine_status{status!="retired",role!="boot"} == 1) by (address) unless sum(chrony_monitor_tracking_system_time_seconds) by (address)
```
zoetrope, line 66: Why is boot excluded?

yokaze: Because the metrics is not scraped from the boot servers.

zoetrope (approved):
LGTM

PR #2044, opened 2021-11-25
add relable_configs for session and sshd

kmdkuk:
- Each start and end of the ssh session was given a unique label value.
- The value of `session-1.scope` etc. at the beginning of ssh is converted to `session.scope`.
- The value of `sshd@231-10.69.0.4:22-10.72.48.0:41850.service` etc. at the end of ssh is converted to `sshd.service`


zoetrope (approved):
LGTM

yokaze (approved):
LGTM

PR #2046, opened 2021-11-25
decrease number of replicas of teleport-proxy to 1

umezawatakeshi:
workaround for recent teleport failure


yokaze (approved):
LGTM

umezawatakeshi:
it is no longer needed because the problem is fixed by https://github.com/cybozu-go/neco-apps/pull/2048 permanently.

PR #2050, opened 2021-11-25
monitoring: add cAdvisor DaemonSet

morimoto-cybozu:


Makefile
```diff
@@ -31,6 +31,17 @@ update-bmc-reverse-proxy:
...
+.PHYNY: update-cadvisor
+update-cadvisor:
+	$(call get-latest-tag,cadvisor)
+	rm -rf /tmp/cadvisor
+	cd /tmp; git clone --depth 1 -b $(call upstream-tag,$(latest_tag)) https://github.com/google/cadvisor
+	rm -rf monitoring/base/cadvisor/upstream/*
+	cp -r /tmp/cadvisor/deploy/kubernetes/base monitoring/base/cadvisor/upstream
+	cp -r /tmp/cadvisor/deploy/kubernetes/overlays monitoring/base/cadvisor/upstream
+	rm -rf /tmp/cadvisor
+	sed -i -E '/newName:.*cadvisor$$/!b;n;s/newTag:.*$$/newTag: $(latest_tag)/' monitoring/base/kustomization.yaml
```
ymmt2005, line 43: why don't we build our own image?

morimoto-cybozu: We actually build and use our own image as you expect.
The name of our image is statically coded in `monitoring/base/kustomization.yaml`.  So it is not the target of this substitution.

I've confirmed that many substitution patterns appear in Makefile.  It's indeed confusing.

ymmt2005: I see.

Makefile
```diff
@@ -31,6 +31,17 @@ update-bmc-reverse-proxy:
...
 	sed -i -E 's,image: quay.io/cybozu/bmc-reverse-proxy:.*$$,image: quay.io/cybozu/bmc-reverse-proxy:$(latest_tag),' bmc-reverse-proxy/base/bmc-reverse-proxy/deployment.yaml
 
+.PHYNY: update-cadvisor
+update-cadvisor:
+	$(call get-latest-tag,cadvisor)
+	rm -rf /tmp/cadvisor
+	cd /tmp; git clone --depth 1 -b $(call upstream-tag,$(latest_tag)) https://github.com/google/cadvisor
+	rm -rf monitoring/base/cadvisor/upstream/*
+	cp -r /tmp/cadvisor/deploy/kubernetes/base monitoring/base/cadvisor/upstream
+	cp -r /tmp/cadvisor/deploy/kubernetes/overlays monitoring/base/cadvisor/upstream
```
masa213f, line 41: Do you use `upstream/overlays`? If it is not used, I think, this copy is unnecessary.

morimoto-cybozu: As you pointed, it isn't used.
I'd like to download it for the convenience of reference from our patch.  Our patch is based on a patch in `upstream/overlays`.
Moreover, I've left the download for some sort of completeness.  This would be a matter of preference.

masa213f (approved):

Makefile
masa213f: I understand. Thank you.

PR #2058, opened 2021-11-29
Stop propagating MetalLB annotation

yokaze:


argocd-config/base/accurate.yaml
```diff
@@ -56,10 +56,7 @@ spec:
             # controller.config.annotationKeys -- Annotations to be propagated to sub-namespaces.
             # It is also possible to specify a glob pattern that can be interpreted by Go's "path.Match" func.
             ## https://pkg.go.dev/path#Match
-            annotationKeys:
-            # An example to propagate an annotation for MetalLB
```
ymmt2005, line 60: Could you fix this in the upstream as well?

yokaze: @ymmt2005 Sure. I would like to write another example, is there any good idea?

yokaze: Possibly `example.com/sample` is one idea

ymmt2005: I'd leave it to you.

yokaze:
This fix is already done in another branch.

PR #2059, opened 2021-11-29
Fix the configuration for ApplicationValidator

zoetrope:
To prevent tenant teams from using `default`, `csa` and `tenant-apps` projects.


yokaze (changes requested):
For this objective, `repository` can be used instead of `repositoryPrefix`.

yokaze (approved):
I added a small fix to this PR. `utility.get_teams` is already a set of unique entries, so there is no need to use `std.set`. Sorry for not mentioning it.

PR #2066, opened 2021-11-30
Rook/Ceph: Expand ceph-ssd osd count.

cupnes:
The number of OSDs in tokyo0 and osaka0 will be set to the same number in stage0.


llamerada-jp (approved):
mobbed

PR #2088, opened 2021-12-08
Prepare Cilium setup

ysksuzuki:
This PR 

- Modifies team-management_test so that it works both with and without Cilium.
- Adds NetworkPolicy to allow hubble-relay to access cilium-agent running on each node.

### DIFF
--- a/test/team-management_test.go
+++ b/test/team-management_test.go
@@ -45,6 +45,9 @@ var requiredResources = []string{
 	"challenges.acme.cert-manager.io",
 	"orders.acme.cert-manager.io",
 
+	// Cilium
+	"ciliumnetworkpolicies.cilium.io",
+
 	// ECK
 	"agents.agent.k8s.elastic.co",
 	"apmservers.apm.k8s.elastic.co",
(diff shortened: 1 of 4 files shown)


test/team-management_test.go
```diff
@@ -190,6 +200,32 @@ func init() {
...
+			if strings.HasPrefix(res, "cilium") {
+				continue
+			}
+			viewableResourcesWithoutCilium = append(viewableResourcesWithoutCilium, res)
+		}
+		viewableResources = viewableResourcesWithoutCilium
+	}
+}
+
+func isCiliumDisabled() bool {
```
Hsn723, line 224: (nit) This function name may be a bit misleading, as based on its content it is checking if cilium is absent, not whether it is disabled. I suggest naming it `isCiliumAbsent` instead.

Hsn723 (commented):
Functionally, I see no problems. My suggestion in the comments is purely cosmetic, plus the functions involved are only there temporarily so I'll leave it up to you to implement it or not.

Hsn723 (dismissed):

zoetrope (dismissed):

Hsn723 (approved):

PR #2095, opened 2021-12-09
Update Grafana Operator to v4.1.1

yokaze:
- Update Grafana Operator to v4.1.1
    - upstream repository has been changed significantly, so manifests are now downloaded from different path.
    - A new CR GrafanaNotificationChannel is added.
      - Since it contains sensitive information, tenant users are allowed to edit it but are not allowed to see other teams' ones.
- Update grafana_plugins_init to v0.0.5
- `preserveUnknownFields` of all the upgraded Grafana Operator CRDs are turned off.
- Invalid fields are removed from Grafana Operator CRs.


monitoring/base/grafana-operator/deployment.yaml
```diff
@@ -1,17 +1,20 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: grafana-operator
+  namespace: system
+  name: grafana-operator-controller-manager
```
zoetrope, line 5: Is this name correct?
The name of the upstream deployment is `controller-manager`.
(Kustomize will add name prefix.)

yokaze: @zoetrope 
Under this PR's coordination, it is correct.
I wanted to add `grafana-operator-` prefix to upstream resources only, so I added an intermediate `kustomization.yaml` under `manifests` directory.

```
- manifests
  - kustomization.yaml (read ../upstream/manifests and add grafana-operator- prefix)
- upstream
- kustomization.yaml
  - read ./manifests
  - patch grafana-operator-controller-manager
```

Another option is to patch `controller-manager` in manifests directory, in that case it applies to `controller-manager` instead of `grafana-operator-controller-manager`.
Which is preferrable?

zoetrope: @yokaze 
I see. It's okay as is.

zoetrope (approved):
LGTM

PR #2107, opened 2021-12-13
fix metallb access test from boot-0

umezawatakeshi:

### DIFF
--- a/test/metallb_test.go
+++ b/test/metallb_test.go
@@ -70,9 +70,9 @@ func testMetalLB() {
 			return nil
 		}).Should(Succeed())
 
-		By("access service from boot-0")
+		By("access service from boot-0 (via cke-localproxy)")
 		Eventually(func() error {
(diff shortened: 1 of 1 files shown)


test/metallb_test.go
```diff
@@ -72,7 +72,7 @@ func testMetalLB() {
 
 		By("access service from boot-0")
 		Eventually(func() error {
-			_, _, err := ExecInNetns("external", "curl", targetIP, "-m", "5")
+			_, _, err := ExecAt("boot-0", "curl", targetIP, "-m", "5")
```
ymmt2005, line 75: On boot-0, services should be accessed with the clsuter IP, not a load balancer IP (`targetIP`).

umezawatakeshi: Oh, right. That was the objective of localproxy.

ymmt2005 (approved):
LGTM

masa213f (approved):

PR #2121, opened 2021-12-15
Relax K8sAPIServersDegraded to 30mins

yokaze:
Currently, `K8sAPIServersDegraded warning` rings on machine reboot, because it takes longer than the alert threshold.
This PR relaxes the time to 30mins, which is equal to that of `NodeExporterDown`.


kmdkuk (approved):
LGTM

PR #2126, opened 2021-12-16
upgrade teleport to 8.0.6

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1772


morimoto-cybozu (approved):
LGTM
Do we need to update the setup steps in Neco Wiki?

umezawatakeshi:
It doesn't seem that the procedure have to be updated.

PR #2130, opened 2021-12-16
Add node failure alert

kmdkuk:
- Add node failure alert
  - Generate an alert labeled `severity: urgent` when `sum(sabakan_machine_status{status="unreachable"}) > 0` for 60 minutes.
  - Confirmed that there are no false alerts due to the reboot of all node.
- Add description about the urgent severity

### DIFF
--- a/monitoring/README.md
+++ b/monitoring/README.md
@@ -53,7 +53,7 @@ We have also added one more severity of our own.
 - `info`: No problem is occurred, but just notify.
 - `warning`: Investigate to decide whether any action is required.
 - `error`: Action is required, but the situation is not so serious at this time.
-- `urgent`: Action is required at best effort because the problem might get worse. Note: `urgent` alerts are intended to be sent to pager only during daytime.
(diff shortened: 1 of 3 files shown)


monitoring/README.md
```diff
@@ -47,11 +47,13 @@ Severity Levels
...
 All alert rules should have the `severity` labels. This label indicates the level of the severity of the alert.
 
-The severity names and their severity order are consistent with syslog severity. We use just four levels from syslog severity, though.
+The severity names and their severity order are mostly consistent with syslog severity. We use four levels from syslog severity, though.
+We have also added one more severity of our own.
 
 - `info`: No problem is occurred, but just notify.
 - `warning`: Investigate to decide whether any action is required.
 - `error`: Action is required, but the situation is not so serious at this time.
+- `urgent`: Action is required at best effort because the problem might get worse. Note: `urgent` alerts are intended to be sent to pager only during daytime.
```
zoetrope, line 56: I think the meaning of "the problem might get worse" is difficult to understand.
For example, how about "the service available, but no redundancy, so we need to take action as soon as possible"?

kmdkuk: Thank you. I've reflected that.

zoetrope (approved):
LGTM

PR #2153, opened 2022-01-05
Set chrony alerts severity

yokaze:


monitoring/base/victoriametrics/rules/node-alertrule.yaml
```diff
@@ -66,15 +66,15 @@ spec:
...
+            severity: error
           annotations:
             summary: The chrony metrics is missing at `{{ $labels.address }}`.
             runbook: Please check the node.
         - alert: NodeClockSkew
           expr: abs(chrony_monitor_tracking_system_time_seconds) > 0.05
           for: 10m
           labels:
-            severity: warning
+            severity: critical
```
kmdkuk, line 77: There is a list of critical alerts in `monitoring/README.md`, so I think it would be better to add them here.
https://github.com/cybozu-go/neco-apps/blob/7a609477a204bb08e77f300db5e44be8e60503fe/monitoring/README.md?plain=1#L59-L100

morimoto-cybozu, line 77: If the difference of 0.05 sec is thought to be critical, i.e. we need to resolve the problem even at midnight, it would be good to add an intermediate warning alert.

monitoring/base/victoriametrics/rules/node-alertrule.yaml
```diff
@@ -66,15 +66,15 @@ spec:
...
           annotations:
             summary: The system clock on `{{ $labels.address }}` is fastened or delayed by 0.05 seconds.
```
morimoto-cybozu, line 79:
```suggestion
            summary: The system clock on `{{ $labels.address }}` is hastened or delayed by 0.05 seconds.
```

kmdkuk (approved):
LGTM

PR #2163, opened 2022-01-07
Fix namespace label on cert-manager metrics.

kmdkuk:
The `namespace=cert-manager` label in the metrics retrieved from cert-manager is meaningless, so change it to be the name of the namespace where the certificate exists.

before metrics

```
certmanager_certificate_*{exported_namespace="<name of namespace where the certificate>", namespace="cert-manager"}
```

after metrics

```
certmanager_certificate_*{exported_namespace="<name of namespace where the certificate>", namespace="<<namespace name where the certificate>>"}
```

The `exported_namespace` label is already in use for alerts and dashboards, so `exported_namespace` will be kept temporarily.
The `exported_namespace` label will be removed when it is no longer used.

- With this change, metrics other than certmanager_certificate_* will also have the new label exported_namaspace=certmanager, which was not there before. However, I think this will not have any side effects as it is only temporary.
- In this PR, I haven't changed the label used for alerts from `exported_namespace` to `namespace`. After applying stage and confirming that the `namespace` label is correctly attached, I will change the alert rules in another PR.

### DIFF
--- a/monitoring/base/grafana-operator/dashboards/cert-manager.yaml
+++ b/monitoring/base/grafana-operator/dashboards/cert-manager.yaml
@@ -104,9 +104,9 @@ spec:
                 ],
                 "targets": [
                     {
-                        "expr": "sort_desc( sum(certmanager_certificate_expiration_timestamp_seconds{exported_namespace=~\"$Namespace\"} - time()) BY (name,exported_namespace) )",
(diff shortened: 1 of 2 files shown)


morimoto-cybozu (approved):
LGTM

PR #2187, opened 2022-01-17
rook: update rook to 1.8.3.0

peng225:
issue: cybozu/csa#71
Signed-off-by: Shinya Hayashi <shinya-hayashi@cybozu.co.jp>


docs/maintenance.md
```diff
@@ -446,59 +446,17 @@ First phase, update Rook solely. Second phase, update Ceph and Rook image based
...
-    ../bin/helm template upstream/chart -f values.yaml -s templates/${i}.yaml > common/${i}.yaml
-  done
-$ for t in hdd ssd object-store; do
-    for i in deployment role rolebinding serviceaccount; do
-      ../bin/helm template upstream/chart -f values.yaml -f values-${t}.yaml -s templates/${i}.yaml --namespace ceph-${t} > ceph-${t}/${i}.yaml
-    done
-    ../bin/helm template upstream/chart -f values.yaml -f values-${t}.yaml -s templates/clusterrolebinding.yaml --namespace ceph-${t} > ceph-${t}/clusterrolebinding/clusterrolebinding.yaml
-  done
+$ rm -rf rook/base/*/charts
+$ make update-rook APP_VERSION=1.8.2.0 CHART_VERSION=1.8.2
```
cupnes, line 455: It's just an example, so maybe you don't need to worry about it.

```suggestion
$ make update-rook APP_VERSION=1.8.3.0 CHART_VERSION=1.8.3
```

test/setup_test.go
```diff
@@ -442,6 +442,34 @@ func applyAndWaitForApplications(commitID string) {
 		}
 	}
 
+	// TODO: remove this block after release the PR bellow
+	// https://github.com/cybozu-go/neco-apps/pull/2187
+	if doUpgrade {
```
satoru-takeuchi, line 447: Just for clarification, I'd like to confirm the purpose of this code block. I confirmed that there are crds of *.openshift.io in the current staging datacenter. These crds seem to be garbage for us but are harmless. In other words, this block is better-to-do than must-do. Is my understanding correct?

peng225: In this PR, CRDs of *.openshift.io are deleted and no longer exist. However, CRDs are not deleted by ArgoCD. So we should do it manually.
It is harmless even if the CRDs are not deleted, but it violates the principle of GitOps. That is why we think we should delete them.

docs/maintenance.md
peng225: This is just an example command, so we would like to leave it as it is.

test/setup_test.go
satoru-takeuchi: OK, got it.

satoru-takeuchi (approved):

cupnes (approved):

PR #2193, opened 2022-01-19
increase vmagent's maxScrapeSize to 200MiB

umezawatakeshi:
increase limit because metrics size of kube-state-metrics significantly increased (from less than 32M to more than 64M).


morimoto-cybozu (approved):
LGTM

PR #2205, opened 2022-01-24
Fix false NodeUnreachable alerts.

kmdkuk:
Fixed false NodeUnreachable alerts when multiple machines were in the `sabakan_machine_status{status="unreachable"}` state alternately.
By changing the evaluation formula to separate instance, address, and serial, an alert will be issued only when one unit is unreachable for 60 minutes.

### DIFF
--- a/monitoring/base/victoriametrics/rules/sabakan-alertrule.yaml
+++ b/monitoring/base/victoriametrics/rules/sabakan-alertrule.yaml
@@ -37,10 +37,10 @@ spec:
             severity: error
         - alert: NodeUnreachable
           annotations:
-            summary: "Unreachable node with a address of {{ $labels.address }} and a serial of {{ $labels.serial }} exist."
+            summary: "Node `{{ $labels.serial }}` at `{{ $labels.address }}` is unreachable."
             runbook: "Perform the server retirement procedure."
           expr: |
-            sum by(instance, address, serial) (sabakan_machine_status{status="unreachable"}) > 0
+            sum by(address, serial) (sabakan_machine_status{status="unreachable"}) > 0
           for: 60m
           labels:
             severity: urgent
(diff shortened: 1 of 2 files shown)


yokaze (changes requested):
Could you please include the instance or address in the message?

kmdkuk:
@yokaze 
Thank you for your comment.
I have modified the summary to show the address and serial that might be used by commands such as `sabactl`.

yokaze (changes requested):

monitoring/base/victoriametrics/rules/sabakan-alertrule.yaml
```diff
@@ -37,10 +37,10 @@ spec:
             severity: error
         - alert: NodeUnreachable
           annotations:
-            summary: "Unreachable node(s) exist."
+            summary: "Unreachable node with a address of {{ $labels.address }} and a serial of {{ $labels.serial }} exist."
             runbook: "Perform the server retirement procedure."
           expr: |
-            sum(sabakan_machine_status{status="unreachable"}) > 0
+            sum by(instance, address, serial) (sabakan_machine_status{status="unreachable"}) > 0
```
morimoto-cybozu, line 43: "instance" is the address of the data source, i.e. the boot server.
I'm not sure whether it is reasonable to sum by "instance."

kmdkuk: Thank you. I think you're right on point.
Currently, there are three instance labels for each address and serial, but the status is synchronized.
Therefore, if an unreachable node appeared, three alerts would be issued when the instance label was split.
I fixed it so that it only sums up the address and serial.

morimoto-cybozu (approved):
LGTM

yokaze (approved):
LGTM

PR #2215, opened 2022-01-27
Fixes retired boot servers so that they do not trigger alerts.

kmdkuk:
Until now, the rule has been that alerts are issued even if the boot server is in a retired state.
The rule has been changed so that alerts will not be issued even if the boot server is in the retired state, since the necessary actions have been taken.
I have also added a test for the alert.


monitoring/base/victoriametrics/rules/sabakan-alertrule.yaml
```diff
@@ -31,7 +31,7 @@ spec:
             summary: "Not-healthy boot server(s) exist."
             runbook: TBD
           expr: |
-            sum(sabakan_machine_status{status!="healthy", role="boot"}) > 0
+            sum(sabakan_machine_status{status!="healthy", status!="retired", role="boot"}) > 0
```
masa213f, line 34: Why don't you exclude retiring?

```suggestion
            sum(sabakan_machine_status{status!~"healthy|retiring|retired", role="boot"}) > 0
```

kmdkuk: I've communicated elsewhere, but I'm writing this for the record.
Since retiring is originally expected to automatically transition to retired, I thought it was an anomaly for the retiring state to continue.
The use of regular expressions to get metrics has been reflected.

masa213f (approved):
LGTM

test/vmalert_test/sabakan.yaml
```diff
@@ -49,6 +49,16 @@ tests:
...
+  - interval: 1m
+    input_series:
+      - series: 'sabakan_machine_status{instance="boot-0", address="1.2.3.4", serial="abcd", status="healthy", role="boot"}'
+        values: 0+0x30
+      - series: 'sabakan_machine_status{instance="boot-0", address="1.2.3.4", serial="abcd", status="retired", role="boot"}'
+        values: 1+0x30
+    alert_rule_test:
+      - eval_time: 30m
+        alertname: BootServerNotHealthy
+        exp_alerts: []
```
morimoto-cybozu, line 61: This test case should be merged into the previous one.

morimoto-cybozu (changes requested):

test/vmalert_test/sabakan.yaml
morimoto-cybozu: I withdraw the previous comment.
The alert expression uses `sum` without preserving dimensions, and we cannot test multiple cases in one set of input_series.

Instead I propose two discrete modifications.
1. Remove the new test case.  We originally do not test all cases.  We do not confirm that the alert is not issued if a boot server is healthy.
2. Use `by(address, serial)` in the alert expression as we do in the `NodeUnreachable` alert.  This would go beyond this PR.

masa213f (approved):
LGTM

morimoto-cybozu (approved):

PR #2220, opened 2022-01-28
Add neco/exclude_noisy_neighbor toreration

kmdkuk:


logging/overlays/gcp/kustomization.yaml
```diff
@@ -30,6 +30,12 @@ patchesJSON6902:
           key: "csa/exclude_noisy_neighbor"
           operator: "Exists"
           effect: "NoExecute"
+      - op: add
+        path: /spec/template/spec/tolerations/-
+        value:
+          key: "neco/exclude_noisy_neighbor"
+          operator: "Exists"
+          effect: "NoSchedule"
```
morimoto-cybozu, line 38: Do we need to specify `effect`?
If the logging-promtail Pods are to be executed even on the nodes with `NoExecute` taints, they should tolerate any type of effects.

toshipp: The CSA team used the NoExecute effect, so I guess you set it up the same way. I agree with your claim. If you remove the effect, please remove NoExecute effects that the CSA team set.

morimoto-cybozu (approved):
LGTM

PR #2224, opened 2022-01-31
Add hubble-ui and hubble-observer

ysksuzuki:
This PR deploys hubble-ui and hubble-observer that collects hubble audit log using hubble-cli. It also adds scrape rules for cilium-agent, cilium-operator and hubble.


Hsn723 (dismissed):

zoetrope (approved):
LGTM

Hsn723 (approved):

team-management/template/neco-admission/base/config.libsonnet
```diff
@@ -85,6 +85,12 @@ function(settings) [{
             'default',
           ],
         },
+        {
+          repository: 'https://helm.cilium.io',
```
Hsn723, line 89: In the light of the recent (albeit fixed) supply chain vuln involving ArgoCD and helm charts, should we really whitelist this third party repository? Although Cilium is a trusted source, I'm worried we might expose ourselves to supply chain attacks in the event this domain gets taken over.

ysksuzuki: Just to be clear, can we still deploy hubble-ui if we don't whitlist the Cilium repository here correct? So what is the effect to add it here?

ysksuzuki: ApplicationValidator only validates repo URLs in Application resource, so it doesn't work in this case.  Retrieving resources from third party repositories and apply them to our cluster is dangerous. So I will store upstream resource with helm template and make argocd apply them with kustomize.

docs/maintenance.md
```diff
@@ -130,6 +131,31 @@ $ make update-external-dns
...
+helm repo update
+helm search repo -l cilium
+# Example output with a header line:
+#   NAME            CHART VERSION   APP VERSION     DESCRIPTION
+#   cilium/cilium   1.11.1          1.11.1          eBPF-based Networking, Security, and Observability
+#   cilium/cilium   1.11.0          1.11.0          eBPF-based Networking, Security, and Observability
+#   cilium/cilium   1.10.7          1.10.7          eBPF-based Networking, Security, and Observability
+#   cilium/cilium   1.10.6          1.10.6          eBPF-based Networking, Security, and Observability
+
+# Choose the same `CHART VERSION` as the version [neco applies]() and set the value like below.
```
Hsn723, line 152: Q: could we simply reference the version in neco's artifacts.go in the release branch? It looks to me like this could be scripted in the Makefile (`update-hubble`). Unless you want to avoid curling artifacts.go from neco.

ysksuzuki: It's a great idea. Thanks!

Hsn723 (approved):

zoetrope (approved):

PR #2262, opened 2022-02-14
Add workaround for switching app with overlays for garoon-apps.

kmdkuk:
Migrate only stage0 to migrate garoon-apps repoURL step by step.
Edit jsonnet directly to put in specific processing.

before(only stage0 overlay): https://github.com/garoon-private/garoon-apps.git
after(only stage0 overlay): https://github.com/cybozu-private/garoon-apps.git 


yokaze (approved):
LGTM

PR #2271, opened 2022-02-15
Replace ApplicationValidator with sourceRepos of AppProject.

zoetrope:


zoetrope:
We will remove ApplicationValidator after all tenants have migrated to cattage,

PR #2290, opened 2022-02-18
remove old sandbox Grafana

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/1871


sandbox/overlays/gcp/dummy-configmap.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
```
masa213f, line 1: Is this dummy resource necessary?
The sandbox applications on tokyo0 and osaka0 are empty.
So I guess, it can be empty on gcp and stage0 as well.

masa213f (changes requested):
Please remove this update command.
https://github.com/cybozu-go/neco-apps/blob/4b20cac5a2c1ed4c5e7bc6ef6ee8a68cd892e548/Makefile#L104

sandbox/overlays/gcp/dummy-configmap.yaml
umezawatakeshi: When an app become empty, `argocd app sync --prune` is required. I don't like to do so each time.

If we decide not to modify sandbox NS anymore, explicit prune may be acceptable.

masa213f (approved):

PR #2300, opened 2022-02-22
remove incomplete storage version migration detection

umezawatakeshi:
This alert rule can only detect migration rarely. It is useless.


kmdkuk (approved):
LGTM

PR #2308, opened 2022-02-25
Fix tests for network policy

ysksuzuki:
This PR fixes the broken network policy tests in which ping failed because the ubuntu pod doesn't have the permission to ping. Also it fixes the domestic egress test.


kmdkuk (approved):
LGTM

PR #2317, opened 2022-02-28
Replace Calico with Cilium NetworkPolicy

ysksuzuki:
This PR includes

- Replace the network policy implementation from Calico to Cilium.
- Remove Calico related files such as the manifest files, scrape and alert configurations, maintenance procedures etc.
- Add the alert configuration for Cilium.

For gcp and stage0, argocd deploys Cilium, for tokyo0 and osaka0 it deploys Calico as before.


Hsn723 (approved):

zoetrope (commented):
Cilium does not have NetworkSet.
It would be easier to understand if you comment on CIDR.

network-policy/base/global-policies/ingress-cluster-allow.yaml
```diff
@@ -1,14 +1,18 @@
...
+      values:
+      - network-nat
   ingress:
-    - action: Allow
-      source:
-        selector: role == 'cluster'
+  - fromEntities:
+    - cluster
+  - fromCIDR:
+    - 10.0.0.0/8
```
zoetrope, line 18: What is the difference between `cluster` and `10.0.0.0/8`?

zoetrope (commented):
I did not understand the need for `sync-wave`.
Does `allow` have to be applied before `deny`?

network-policy/base/global-policies/ingress-cluster-allow.yaml
ysksuzuki: `cluster` is the in-cluster endpoints such as pods, nodes inside the cluster. `10.0.0.0/8` indicates the endpoints outside the cluster such as boot servers, out-cluster nodes. We need allow `10.0.0.0/8` as well, otherwise those can't communicate with pods.

ysksuzuki:
> I did not understand the need for `sync-wave`. Does `allow` have to be applied before `deny`?

I think we don't need to specify sync-wave with CiliumNetworkPolicy. It's just the sync-wave specified in CalicoNetworkPolicy.

network-policy/base/policies/domestic-egress.yaml
```diff
@@ -1,48 +1,32 @@
...
+    - ports:
+      - port: "8080"
+        protocol: TCP
   egress:
-    - action: Allow
-      destination:
-        selector: role == 'switch'
-    
+  - toCIDR:
+    # Switch network
```
Hsn723, line 29: Purely cosmetic, but instead of using comments it might be more elegant to use `.spec.description` to write a slightly more detailed description. Of course, some policies are pretty self-explanatory and may not warrant a full explanation, but this could be useful for more complex policies (plus it has the advantage of being `kubectl`-able)

Hsn723 (approved):

zoetrope (approved):
LGTM

PR #2326, opened 2022-03-01
Add libsonnet to generate Tenant resources

zoetrope:
This is just a template. It does not generate anything now.


yokaze (changes requested):
I would like to discuss the following points.

1. All the files related to a tenant should be cleaned up after it is removed.
  In this case, `team-management/base/tenant-<tenant>.yaml` remains after the deletion of the tenant. This should be handled in `make team-management`.

2. I would like to keep the generated manifests unchanged as possible when we modify `settings.json`.
  It looks good to sort the following arrays:
  - https://github.com/cybozu-go/neco-apps/blob/ab5a254444ebc00078685e2a56af531f852ccaa1/team-management/template/neco-admission/config.libsonnet#L38
  - https://github.com/cybozu-go/neco-apps/blob/ab5a254444ebc00078685e2a56af531f852ccaa1/team-management/template/teleport/base/nodes/serviceaccount.libsonnet#L11

3. `get_tenants` function is added between `get_teams` and `get_team_namespaces`. I would like to place it before or after the two functions because they are correlated. 

4. I personally added `template` to all the libsonnet files that actually generates files of different names. This eases to investigate the template file tree. So I would like to rename `tenant.libsonnet` to `tenant-template.libsonnet`.

5. Two types of privileges will be removed if a tenant is moved to the `"tenant"` section.
  - The team will not be able to get and sync `tenant-app-of-apps` via Argo CD dashboard. It will be a concern if a tenant team wants to deploy their own operators or cluster resources, but looks non-problematic at this moment.
  - We will not be able to use `maneki` AppProject to deploy resources to the namespaces maintained by Tenant resources. Since the AppProject is used very widely, it will be difficult for the SRE team to migrate their namespaces without a mitigation method.
  This point can be postponed for a while.

zoetrope:
@yokaze 
> I personally added template to all the libsonnet files that actually generates files of different names.

I could not find it.
Which one is it?

zoetrope:
@yokaze 
> The team will not be able to get and sync tenant-app-of-apps via Argo CD dashboard. It will be a concern if a tenant team wants to deploy their own operators or cluster resources, but looks non-problematic at this moment.

`tenant-app-of-apps` is not `tenant-apps`.
Tenant teams has `tenant-apps` permissions and can deploy their operator.

zoetrope:
@yokaze 
> We will not be able to use maneki AppProject to deploy resources to the namespaces maintained by Tenant resources. Since the AppProject is used very widely, it will be difficult for the SRE team to migrate their namespaces without a mitigation method.

Cattage provides a migration method.
If ownership of a root-namespace is transferred between tenants, AppProject will be adjusted automatically.

yokaze:
@zoetrope 
Examples for files named template are:
- `argocd-config/overlays/template`
- `argocd-config/overlays/template/tenants/template-apps.libsonnet`
- `team-management/base/team-template`

`teleport/base/conf/team-role.libsonnet` does not include template, but it is perhaps able to guess that it generates files of different names. Strictly speaking it is better for this file name to include template.

yokaze:
@zoetrope 
> `tenant-app-of-apps` is not `tenant-apps`.
Tenant teams has tenant-apps permissions and can deploy their operator.

Thanks for clarification. To my understanding tenant teams will lose permission to get and sync `cybozu-private/neco-tenant-apps`, is it intended?

zoetrope:
@yokaze 
> Thanks for clarification. To my understanding tenant teams will lose permission to get and sync cybozu-private/neco-tenant-apps, is it intended?

No.
Tenant teams still have the permission.

yokaze (approved):
LGTM!

PR #2340, opened 2022-03-04
add app-csa-privileged NS with privileged policy

umezawatakeshi:
see internal request NECOREQ-32


kmdkuk (approved):
LGTM

PR #2341, opened 2022-03-04
add argocd-notifications

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/778


secrets/base/argocd/secret.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: argocd-notifications-secret
```
morimoto-cybozu, line 4: Don't we need `namespace: argocd`?

If argocd-notifications can run without this secret, i.e. if the deployment does not mount the secret at start-up but just refers to it dynamically, this secret may be omissible.
Of course it would be better to provide this secret, though.

umezawatakeshi: Yes, we need namespace. 👍

morimoto-cybozu (approved):
LGTM

PR #2351, opened 2022-03-07
fix argocd-notifications subscription selector

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/778


kmdkuk (approved):
LGTM

PR #2354, opened 2022-03-07
Set tolerations to DaemonSet pods against CKE's taints

masa213f:
https://github.com/cybozu-go/neco/issues/1941


topolvm/base/values.yaml
```diff
@@ -58,10 +63,12 @@ node:
   
   # Temporary configuration for storage performance measurement
```
umezawatakeshi, line 64: This comment have to be moved because it does not cover `cke.cybozu.com/state` toleration.

umezawatakeshi: line 42 too.

umezawatakeshi (approved):

llamerada-jp (approved):

PR #2396, opened 2022-03-18
fix bad indent of line continuation

umezawatakeshi:
It may not be a problem for the parser since it is single-quoted, but VSCode redline it.


morimoto-cybozu (approved):
LGTM

PR #2397, opened 2022-03-18
Upgrade test VM to n2-standard-96

morimoto-cybozu:


morimoto-cybozu:
Close this because we are now trying n2-standard-80.

PR #2402, opened 2022-03-22
add loki and promtail with TopoLVM backend

cupnes:
issue: cybozu/csa#59


Makefile
```diff
@@ -198,6 +200,30 @@ update-logging-promtail:
 	$(call get-latest-tag,promtail)
 	sed -i -E '/name:.*promtail$$/!b;n;s/newTag:.*$$/newTag: $(latest_tag)/' logging/base/promtail/kustomization.yaml
 
+CEPH_LOKI_IMAGE_VERSION := $(shell awk '/LOKI_IMAGE:/ {print $$2}' logging-small/base/VERSIONS)
```
toshipp, line 203: nits: this loki is indeed for ceph, but we call this is logging-small, so how about calling it `SMALL_LOKI` ?

toshipp (changes requested):

toshipp (approved):

PR #2412, opened 2022-03-24
Introduce templates for tenant network policies

Hsn723:


tenet-policy/base/admission/forbid-egress-node.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: tenet.cybozu.io/v1beta2
+kind: NetworkPolicyAdmissionRule
+metadata:
+    name: forbid-egress-node
+spec:
+    namespaceSelector:
+      excludeLabels:
+        team: neco
+    forbiddenIPRanges:
```
ysksuzuki, line 9: Tenants users can allow egress to the ports on nodes which neco's policy doesn't allow by allowing host/remote-node entity. Do you think that it is too much if we prohibit tenant users from allowing those ports on the nodes using the admission webhook?

ysksuzuki:
Can you add tests to verify this implementation? The specs we need to verify would be the following if I understand it correctly.

 - The new implementation won't break the current netpol specification. 
   - Pods inside a tenant namespace can communicate with Pods outside the namespace.
 - Tenant users can opt-in the default ingress deny of communication across namespaces properly.
 - Tenant users can't allow node and bmc CIDR.

tenet-policy/base/admission/forbid-egress-node.yaml
Hsn723: Unfortunately tenet currently doesn't support entity or port based admission rules, although that could be something to look into.

Hsn723:
> Can you add tests to verify this implementation? The specs we need to verify would be the following if I understand it correctly.
> 
> * The new implementation won't break the current netpol specification.
>   
>   * Pods inside a tenant namespace can communicate with Pods outside the namespace.
> * Tenant users can opt-in the default ingress deny of communication across namespaces properly.
> * Tenant users can't allow node and bmc CIDR.

From my understanding of the network-policy_test.go code, `Pods inside a tenant namespace can communicate with Pods outside the namespace` is already being handled there, so only the last two conditions need to be tested?

Hsn723:
Also, due to the presence of `tenant-ingress-cluster-allow` (during the migration period), testing `Tenant users can opt-in the default ingress deny of communication across namespaces properly` would be difficult without being able to overwrite `tenant-ingress-cluster-allow` somehow. Would temporarily disabling sync for the network-policy Application in ArgoCD be acceptable?

ysksuzuki:
> Pods inside a tenant namespace can communicate with Pods outside the namespace is already being handled there

AFAIK, we haven't checked communication between tenant namespaces.

> Tenant users can opt-in the default ingress deny of communication across namespaces properly would be difficult without being able to overwrite tenant-ingress-cluster-allow somehow.

Maybe we should verify this when tenants actually opt-in. Disabling sync and override a policy is bit much.

zoetrope (approved):
LGTM

PR #2415, opened 2022-03-24
loosen PersistentVolumeUsageRapidIncrease alert for VictoriaMetrics PVs

umezawatakeshi:
The usage of VictoriaMetrics PVs is saw-toothed. The original alert rule is too sensitive and causes many false positives.


morimoto-cybozu (approved):
LGTM

morimoto-cybozu (approved):
LGTM

PR #2430, opened 2022-03-29
The status of app-of-apps application must always be HEALTHY

zoetrope:
Tenant users can use cattage to manage app-of-apps application resources themselves.
In order to synchronize app-of-apps application resources, their status must always be HEALTHY.


yokaze (approved):
LGTM

PR #2449, opened 2022-03-31
improve node unreachable alerts

umezawatakeshi:
- `BootServerNotHealthy`: fire individual alerts for each boot server.
- `NodeUnreachable`: output rack number in alerts.


morimoto-cybozu (approved):
LGTM

PR #2456, opened 2022-04-01
Tenant migration mode

zoetrope:
When migrating to tenants, some teams will want to migrate only stage and dctest.

Set the team name to `migrating_teams` in settings.json and remove `stage0` from `apps.<team>.destinations`.
As a result, tokyo0 and osaka0 will keep current manifests, stage0 and dctest will be migrate to tenant.


yokaze (approved):
LGTM for a limited time use, but it looks a little bit hard to read. It is better to overhaul the team-management jsonnet.

PR #2514, opened 2022-04-21
Replace a URL for the node-exporter dashboard.

kmdkuk:
When importing a dashboard using the `.spec.url` of `GrafanaDashboard`, a 429 error often occurs, so I replaced it with a url that was hosted independently on gcloud storage.


masa213f (approved):
LGTM

PR #2589, opened 2022-05-12
Add user and token for cloning neco-apps in dctest-runner

kmdkuk:

### DIFF
--- a/meows/overlays/stage0/runnerpool.yaml
+++ b/meows/overlays/stage0/runnerpool.yaml
@@ -27,7 +27,7 @@ spec:
         egress.coil.cybozu.com/internet-egress: nat
     automountServiceAccountToken: false
     runnerContainer:
-      image: quay.io/cybozu/meows-dctest-runner:0.7.0.1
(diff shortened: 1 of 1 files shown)


meows/overlays/stage0/runnerpool.yaml
```diff
@@ -27,14 +27,21 @@ spec:
...
+      image: quay.io/cybozu/meows-dctest-runner:0.7.0.3
       securityContext:
         privileged: true
       env:
       - name: NECO_BRANCH
         value: "release"
       - name: NECO_APPS_BRANCH
         value: "release"
+      - name: GITHUB_USER
+        value: cybozu-neco
```
masa213f, line 39: In my sence, the user name is a sensitive data.
I think it is better to set `GITHUB_USER` via a secret as well.

But this name is already hard coded in neco-apps.
So in this case, it’s fine as it is.

masa213f (changes requested):

masa213f (dismissed):
LGTM

masa213f (approved):

PR #2601, opened 2022-05-17
the commands fails, if cybozu-private-repo-read-pat does not exist.

kmdkuk:


morimoto-cybozu (approved):
LGTM

